PR #1, opened 2017-05-15
App setup (pt1 - visual identity and description)

shaunanoordin:
## PR Overview
* This PR is intended to set up the **app's identity**, distinct from the starter template it's using.
  * The app's name and description has been updated across the app itself, the README, etc.
  * The app's theme is being set up in `app-theme.styl`
* FYI it's supposed to look like this:
screen shot 2017-05-15 at 18 35 14
* And it looks like this right now:
screen shot 2017-05-15 at 18 35 30
* It's... it's a work in progress.
* Of note:
  * We need to get those fonts correct.
  * **Is there a common Zooniverse Header we should use?**
* OK, there are also some small personal tweaks, like renaming site-header to app-header, because I like consistency like that.

### Status
Ready for merge; this is just a small commit to get things started.

@wgranger we can discuss moving forward from here - I want to spend the rest of the week nailing the design provided by Becky, but if we know the site functionality it may also make sense to start 'partitioning' the app into individual components we can work on.


shaunanoordin:
## PR Update
* OpenSans and LeagueMono fonts added.
* The .JS files are renamed to .JSX, which make more sense.

wgranger:
It looks like there's a Zooniverse-React-Components dependency we've been using in lab-app for the logos, etc., so we should consider introducing that and also modifying parts of the package.json a bit.

wgranger:
I'm not sure what is going on with some of these styles not being visible. My view is attached, and it differs from the image above. 
screen shot 2017-05-18 at 3 22 01 pm

shaunanoordin:
@wgranger Whoops, sorry, I forgot to update my first screenshot - your screenshot is correct, as that's how Pt 1 looks like now. I had spun off the _nicer-looking_ app header/navi into Pt 2.

shaunanoordin:
Zooniverse-React-Components = 👍 

I'd recommend we do that in PR #2 though, unless there are major things to change in this PR.

wgranger:
Great. I was wondering why it looked so different. I think this branch is good then.

PR #5, opened 2017-06-01
Add React Components to package.json

wgranger:
Simple as it looks


srallen:
@shaunanoordin We just haven't added the header and footer to it yet. It's definitely on the list to do.

PR #6, opened 2017-06-02
Shared Zooniverse header [placeholder?]

shaunanoordin:
## PR Overview
* This PR adds the shared Zooniverse header to the project.
screen shot 2017-06-02 at 15 39 06
* **Caveat:** this isn't the "real" Zooniverse header, just a placeholder.
  * We're expecting a new Zooniverse shared header from the Zooniverse-React-Components, but we don't have an ETC at the moment.
  * One option is to create a functional shared Zooniverse header specific for this project, then simply replace it when the 'real' header from the ZRC is confirmed.
* In any case, this PR also updates `app-themes.styl` to closer reflect the updated designs on Zeplin from the past 24 hours.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Header } from './Header';
+import Header from './Header';
+import ZooniverseHeader from './ZooniverseHeader';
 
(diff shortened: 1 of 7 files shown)


shaunanoordin:
This will be superseded by a new style-update PR - coming soon to a repo near you!

PR #7, opened 2017-06-05
Design v3 style update

shaunanoordin:
## PR Overview
* This PR updates the visual look (colours, fonts, layout, etc) to match the latest (v3) design guide.
* This is mostly a design placeholder; functionality is not being built yet.
* We are still missing a shared Zooniverse header, but this should be coming from the Zooniverse-React-Components library.

### Visual Reference
screen shot 2017-06-05 at 16 49 10
https://indd.adobe.com/view/e2b8cf36-eb64-4e7b-8cf0-c923214d2e67

### DIFF
--- a/src/containers/ClassifierContainer.jsx
+++ b/src/containers/ClassifierContainer.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 //import { checkLoginUser, loginToPanoptes, logoutFromPanoptes } from '../ducks/login';
 
 import TmpSvgIcon from '../components/TmpSvgIcon';
+import TmpImg from '../images/zooniverse-icon-web-white-small.png';
 
 class ClassifierContainer extends React.Component {
   constructor(props) {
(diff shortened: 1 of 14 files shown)


shaunanoordin:
* We now have a placeholder `/classify` page, and this is how it looks like:
screen shot 2017-06-05 at 17 37 06

shaunanoordin:
Update:
screen shot 2017-06-07 at 16 37 45

shaunanoordin:
OK we're done here.

screen shot 2017-06-09 at 13 34 26

Ready for review!

src/containers/ClassifierContainer.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+//import { checkLoginUser, loginToPanoptes, logoutFromPanoptes } from '../ducks/login';
```
wgranger: I take it these are the ducks that still need to be written for classifier functionality.

wgranger:
This looks good overall and is quite succinct. A good bit was done with a few lines. The next step seems to be getting those ducks in order for api calls. Something we need to think of in the future is, again, making it mobile-friendly, but that is not a concern at this point and can prob be done in a separate PR later with a few lines of code. Overall, this provides a nice placeholder for components to come.

I like the use of items like `rem` and `<i>` in this PR - something we don't use too often in PFE. 

I've tested this out in all browsers, and everything appears to work well. I was unable to get IE up and running on virtual box, but that shouldn't hold up this PR.

src/containers/ClassifierContainer.jsx
shaunanoordin: Quack, quack quack quack. (Yes, that is correct.)

Issue #10, opened 2017-06-09
Connect initial ducks for classifier page


wgranger:
Currently, three are commented out:

checkLoginUser
loginToPanoptes
logoutFromPanoptes

More will be added as the app grows.

Issue #8, opened 2017-06-08
Create the Classifier


wgranger:
At its core, this classifier will work similarly to the one on PFE without all the experimental bells and whistles.

shaunanoordin:
The Classifier is pretty straightforward, but we can break this down into smaller independent tasks.

### Task 1: Fetch Subjects from Panoptes
* Goal: using the API client, get a random subject (as JSON) from Panoptes 
  * e.g. `panoptesApiClient.type('subjects').get('671189').then(...);`
* **Caveat:** For testing purposes, we'll need to _either..._
  * set up a new LtL project on `staging` (more setup work) or 
  * use `production`'s more "true-to-life" AnnoTate data (we need to switch our dev environment to production).

### Task 2: Display Subject
* Goals:
  * Given some subject data (ID, image URL, metadata), show the Subject (e.g. a journal page) visually on the web page.  
  * Allow the user to zoom/rotate/move the Subject.
* This will mostly involve SVG magic hooked up with keyboard/mouse/touch events.
  * I'll be using this for reference: https://github.com/shaunanoordin/zooniverse-react-transcriber
* For bonus points: keep ACCESSIBILITY in mind. Keyboard shortcuts FTW!

### Task 3: Send Classifications to Panoptes
* Goal: given some annotations, submit the Classification back to Panoptes
* In the future, we may also need to add extra metadata to the Classifications, e.g. the number of times a user accessed a collaborative feature. Don't worry about this for now, but keep it in mind.
* How would the annotations/Classifications be structured? I assume it'll be spelled out in the workflow, but this is worth checking.
* **Caveat:** we'll need to eventually check if the data we're feeding back is legit (valid, usable, and aggregatable) but this may be a bigger issue than just the Classifier.

### Task 4: Allow users to make Annotations
* Goal: allow users to mark start coordinates and end coordinates on a Subject, then enter some text.
* A corollary to Task 2 and Task 3.

### Task 0: Include Panoptes Javascript Client 
* Basically: https://github.com/zooniverse/panoptes-javascript-client
* We add `panoptes-client` `2.7.2` or [whatever](https://www.npmjs.com/package/panoptes-client) to `package.json`
* This task will take approximately 30 man hours and about 4 weeks of regression testing.

### Other
* Features such as a collaborative dropdown should be separated into their own issues. Let's keep this fairly simple.

srallen:
[...]

PR #11, opened 2017-06-14
Panoptes hookup - fetch Project data and Subjects for Classifier

shaunanoordin:
## PR Overview
* In this PR, we're hooking up the custom front end with the Panoptes API.
* I'm focusing on the following implementations:
  * The CFE should be able to fetch project details from Panoptes
    * For example, staging project 1651, aka http://panoptes-staging-url/projects/darkeshard/transformers
  * The CFE should be able to fetch the queued subject for the current Zooniverse user looking at that specific project.
* See Issue #8 and Issue #10 for further spec details.
  * Specifically, look at [Task 1: Fetch Subjects from Panoptes](https://github.com/zooniverse/liberating-the-liberator/issues/8#issuecomment-307826733)
  * Also note [Sarah's comments](https://github.com/zooniverse/liberating-the-liberator/issues/8#issuecomment-307833120) on the subject queue endpoint.

Random thoughts:
* [Taxanomy? Ontology?] Right now the code's sitting in `src/ducks/classifier.js`, but I'm wondering if it should logically go into `src/ducks/projects.js` and `src/ducks/subjects.js`... hmmm.


shaunanoordin:
Actually, now that I think about it, I really should leave the `ClassifierContainer` component alone. I'll set up a new `ClassifierTester` component where we can more freely play around with the duck functionality and the Panoptes hooks without worrying about presentation.

wgranger:
I think it makes more sense to separate these items into two ducks, `projects.js` and `subjects.js`. It'll be a lot more cleaner than potentially having a monolith `classifier.js`.

shaunanoordin:
Update: I've created a simple `classifier-tester` that's viewable at http://localhost:3000/classify-tester . It does one very simple thing: fetch & display the details of staging Project 1651.

We can use this as the foundation for future Panoptes API data testing.

shaunanoordin:
@wgranger I've added some extended comments for the Project duck and the Classify (Panoptes API) Tester; let me know if you have concerns about the code being too wordy/not wordy enough.

Ready for review!

wgranger:
Travis hates that no environmental variables are set.

src/containers/ClassifierTester.jsx
```diff
@@ -0,0 +1,100 @@
...
+class ClassifierTester extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <main className="app-content tester-page">
+        <div className="control-panel">
+          <button className="green button" onClick={this.fetchProject.bind(this)}>Fetch Project</button>
```
wgranger: can this bind be put in the constructor? `this.fetchProject = this.fetchProject.bind(this)`

wgranger:
Overall, this PR looks good and it ready to merge once the Travis issues and line comment are addressed.

shaunanoordin:
Awright, I've added some updates:
* The event bind for `this.fetchProject()` in `ClassifierTester` has been moved to the constructor.
* Some updates to the Travis configuration have been made.

I don't think Travis has an issue with no environmental variables being set - I believe Travis is doing a default check using yarn (which isn't properly set up for our project) instead of our custom test script.

shaunanoordin:
UPDATE: OK so the Travis problem may be affecting more people than just us.

1. Check the Slack conversation on #lab-app on 15 June 2017. Of note:
```
Sarah
[15:17] @here this may be the cause of the issues we're having with travis: https://kb.heroku.com/why-is-my-node-js-build-failing-because-of-conflicting-lock-files
[15:18] We should try to switch back to npm 5 managing packages, remove the yarn.lock file and see if travis will handle it
```

2. Also check out https://github.com/travis-ci/travis-ci/issues/7119 - it states how Travis tries to install yarn by default. (And apparently we have to... uh... overwrite npm install? Or something?)

True enough, if we check one of the job logs in this PR - e.g. https://travis-ci.org/zooniverse/liberating-the-liberator/jobs/242927829 - we'll see errors along the likes of `The command "eval yarn " failed. Retrying, 2 of 3.`

Heck if I know what's going on with that darn yarn.

wgranger:
Seems like we still have some things to figure out with Travis. However, this shouldn't block this PR as taking the branch for a test run worked well. I'm going to merge, and we can figure out Travis separately.

shaunanoordin:
Sweet - we'll open an investigation in Travis!

PR #14, opened 2017-06-19
Retrieve Subject and Queue

wgranger:
This is based pretty closely to the project retrieval with the addition of the queue. There might be some intrinsic changes that need to be made, as this is the second time I've used Redux, so I'm open to suggestions.

Working through this, there should also be a subject location reader helper method, similar to what exists in PFE. I might push this up in another commit, but if you review this before that commit, feel free to merge and I'll add the addition in another PR.

### DIFF
--- a/src/containers/ClassifierTester.jsx
+++ b/src/containers/ClassifierTester.jsx
@@ -11,11 +11,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { fetchProject, PROJECT_STATUS } from '../ducks/project';
+import { fetchSubject, SUBJECT_STATUS } from '../ducks/subject';
 
 class ClassifierTester extends React.Component {
   constructor(props) {
     super(props);
     this.fetchProject = this.fetchProject.bind(this);
+    this.fetchSubject = this.fetchSubject.bind(this);
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


wgranger:
Also, the Yarn error appears similar to what you were experiencing in a recent PR, so the Yarn issue is still outstanding.

wgranger:
Ok, I wrote a little helper method and then realized afterwards we might not have any multi-image subjects! However, this should also work well to check images contain compatible extensions and also make it cleaner to get the source of a subject. There could possibly be a more succinct way to write that method.

shaunanoordin:
I've temporarily disabled Travis checks on `liberating-the-liberator` while we figure out the yarn issues.

PR checks are ongoing, but what I'm seeing so far LGTM! The only thing I'm still digesting is the structure of our helper methods.

simoneduca:
@shaunanoordin I think this might be relevant https://kb.heroku.com/why-is-my-node-js-build-failing-because-of-conflicting-lock-files
As per a Slack convo, the issues people were facing with Travis were caused by running `yarn` and `npm` commands, which generated confusion in Travis. Also I believe we're actually moving away from Yarn. SO I'd say it's safe to remove `yarn.lock` and add a `package-lock.json`. I've done so in Pandora and Travis is running fine.

srallen:
Yarn is still really new and basically npm implemented in version 5 one of the major differences, lock files, between it and npm . We never really had enough time to implement yarn for a project that's made it to production. I'm of the opinion that we should drop our experimenting with yarn since npm can pretty much do what what yarn does now and is more stable with existing feature sets (yarn was buggy with many standard npm features).

PR #17, opened 2017-07-07
Update Classifier Design

wgranger:
This is an update to the new classifier prototype. I made a good bit of changes here, and even started on a global, Zooniverse header, which still needs a bit of work. I took some liberties with design and placement of files, so things are up for debate.


shaunanoordin:
## PR Overview
* This PR is primarily a visual style update for project, notably for the Classifier.
screen shot 2017-07-11 at 15 02 07
_Classifier page_
screen shot 2017-07-11 at 15 02 18
_Home page_
* Generally, the PR looks good and no further changes are requested before approval, but I have some discussion points.

### Points of Discussion
* The newer CSS class names follow a pattern of `main-component__child-component--state`, using double underscores and double dashes, e.g. `app-header__nav-links` and `project-header__link--active `
  * Out of curiosity, is there a rationale behind this naming convention? e.g., versus a modular naming convention like `<div class="panel"><button class="active red button" /></div>` and ` .active.button { ... } .red.button { ... } .panel .button { ... }`
* I'm not sure `TOOLBAR_CONTROLS` makes sense as a repeatable set of items (like the `SiteNavItems`) since each Toolbar item tends to be unique.
  * For example, we'd need to add different `onClick` handlers for each of them (`this.rotateLeft`, `this.zoomIn`) and each might have unique logic states (e.g. the Contrast button should change to indicate it's switched on/off)
  * Essentially, I don't see the advantage of putting the Toolbar Controls into a 'mappable array', as opposed to a bunch of individual `<buttons />`/etc. Do you have thoughts on this?

### Status
Approved and merging. Tagging @wgranger for the discussion points, above.

PR #19, opened 2017-07-14
Add SubjectViewer

shaunanoordin:
## PR Overview
* Adds a basic Subject Viewer component to the Classifier page.
  * The Subject Viewer will display a single Subject image (using SVGImage)
  * Currently, a dummy image from the AnnoTate project is being used.
  * A user can navigate the Subject image by using mouse controls:
    * Click-drag to pan
    * Mouse wheel to zoom
  * **NOTE: The SubjectViewer and SVGImage both set the (0,0) origin to the CENTRE of the SVG**
* The Classifier page has been updated to add functionality to certain control buttons.
  * Users can Zoom In, Zoom Out, Rotate the image, Reset the image, and switch between Annotation mode and Navigation mode. (Note: Annotation mode doesn't do anything at the moment.)
* A Subject Viewer Redux Duck has been added to control the navigation/transformations/view options of the component. (i.e. zoom values, rotation values, etc)

### Notes
* There may be some glitchiness to the sizing of the SVG,
* ...and this component still needs to be tested across all browsers. Including That Which Must Not Be Named 11.
* For the moment, as long as this works on Chrome, it should be sufficient for the needs of the Wednesday meeting.

### Status
Ready for review, assigning to Will. (Note: I think with the new PFE standards, we're asking the _Assignee_ to not only _approve,_ but to also initiate the _merge._ )


shaunanoordin:
## PR Update
screen shot 2017-07-17 at 18 24 57
_Current progress of the Classifier page, viewed on OSX+Chrome_

* We've added a new SubjectViewer component to the Classifier page, whose job is to display the Subject in question.
  * The SubjectViewer correctly reads mouse input to pan and/or zoom in on the Subject.
  * Currently, we've hooked up a temporary Subject from AnnoTate.
* There's a new Redux Duck that keeps track of the rotation/scaling/translation of the Subject on the Subject Viewer.
* The Classifier page's Toolbar controls have minor functionality added to them, where possible.

src/components/SVGImage.jsx
```diff
@@ -0,0 +1,69 @@
...
+  
+  render() {
+    if (this.state.loaded) {
+      return (
+        <image className="svg-image"
+          href={this.image.src}
+          width={this.image.width}
+          height={this.image.height}
+          x={(this.image.width * -0.5)+'px'}
+          y={(this.image.height * -0.5)+'px'} />
```
wgranger: This is probably related to what we discussed yesterday. I'm guessing you are multiplying the width and height by -0.5 to move the image to the top left corner of the container.

src/containers/SubjectViewer.jsx
```diff
@@ -0,0 +1,252 @@
+import React, { PropTypes } from 'react';
```
wgranger: This should probably pull `PropTypes` from the `prop-types` package instead of `'react'`.

src/containers/SubjectViewer.jsx
```diff
@@ -0,0 +1,252 @@
...
+  ACTIVE: 1,
+}
+
+class SubjectViewer extends React.Component {
+  constructor(props) {
+    super(props);
+    
+    //HTML element refs.
+    this.section = undefined;
+    this.svg = undefined;
```
wgranger: Is it necessary to declare the refs as `undefined` initially? It seems a bit odd to me, and I was under the assumption that refs would be initiated once declared within the element.

src/containers/SubjectViewer.jsx
```diff
@@ -0,0 +1,252 @@
...
+      </section>
+    );
+  }
+  
+  //----------------------------------------------------------------
+  
+  componentDidMount() {
+    //Make sure we monitor visible size of Subject Viewer.
+    window.addEventListener('resize', this.updateSize);
+    this.updateSize();
```
wgranger: I also added an event listener to `resize` in the new homepage PR to resize the background image. I doubt it will affect this component, but it's worth double checking.

src/components/SVGImage.jsx
shaunanoordin: Yup, it's a centre-ing compensator. Might be worth updating this with some comments to explain that.

src/containers/SubjectViewer.jsx
shaunanoordin: This should **not** be a problem as long as:
1. neither of us uses `window.onResize = ...` and instead use `window.addEventListener()`
2. we both use `window.removeEventListener` when the component unmounts, and this triggers properly.

src/containers/SubjectViewer.jsx
shaunanoordin: I generally practise declaring refs in the constructor so people know what to expect later on in the code. This isn't technically essential, but I'd rather be explicit in spelling out the parts/elements/members/dangly bits of any component.

eatyourgreens: I think a better default for DOM nodes or references to components is `null` instead of `undefined`.

eatyourgreens: 👍  to being up front about declaring properties.

src/containers/SubjectViewer.jsx
wgranger: Those should both be the case, so I think we're in the clear.

shaunanoordin:
## PR Update
* This PR is now ready for review and out of the WIP status.
* The initial (blank) PR description has been updated with the goals/intent of this PR. Please scroll to the top for full details.

### Status
Ready for review

wgranger:
Overall, this looks great. Only two items of note:

- When panning, should we change the cursor from an arrow to something else? Intersecting arrows?

- When I reset the image, it doesn't return to the original image size as seen on first load. Instead, it reverts to a zoomed-in version.

I'm happy to merge as is and we can address these issues later, unless you believe this will be a quick fix.

shaunanoordin:
Yipes! I forgot about the Reset Image button and how it'd interplay with the best-fit feature. Thanks for catching that.

Also, good point on the mouse cursors - I'll need to add this to make the UI more sensible.

Unfortunately, both these items will require a fair amount of fixing on the backend, so I'm going to mark these as issues and approach them tomorrow/next week.

wgranger:
Perfect. I'll go ahead and merge before heading out today.

PR #20, opened 2017-07-17
Homepage Redesign

wgranger:
This updates the homepage to the new prototype.


src/components/Home.jsx
```diff
@@ -1,14 +1,85 @@
...
+      return <div key={i}>Topic {num} &#8608;</div>;
+    });
+  }
+
   render() {
     return (
       <main className="app-content home-page">
-        <b>Welcome to Liberating the Liberator.</b> Now let&apos;s drop some of that phat Lorem ipsum beat.
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eget eleifend risus, id aliquet est. Maecenas tempus luctus lacinia.
+        <div className="project-background" style={{ height: this.state.backgroundHeight }} />
```
shaunanoordin: I'm trying to wrap my head around the need for having a distinct `.project-background` element, one that has its height calculated dynamically - I assume that a standard CSS `background: url(), background-size: cover` on the `home-page` element wouldn't work? Let me poke around the code a bit before I comment further.

shaunanoordin:
screen shot 2017-07-18 at 12 42 57

Weird bits of derp here.
* On Chrome, when you _first load the home page_ at _specific widths_ (e.g. 1200px <= width <= 1300px) there's a white gap between the project background, as the project-background element doesn't calculate the height properly.
  * I'm wondering if it may simply make more sense to set `.project-background`to `height: 90vh` or similar
* There's also a weird white gap after the end of the page, on the home page.

src/components/Home.jsx
wgranger: Yes, I think there might be a better way to have the project background included. This is what I was referring to earlier when I said there might be some messy code in this PR. In short, I'm trying to dynamically change the height of the background so something like the white line you mentioned below doesn't appear on resize.

srallen: Just to echo what @shaunanoordin said, I think `background-size: cover` should be all that you need, with the url also set in the CSS on the body rather than having a separate div just for this.

shaunanoordin:
## PR Review
* This PR updates the homepage to match Becky's latest designs for the (Abolitonist Papers aka Liberating the Liberators aka) Anti-Slavery Manuscripts project.
* Most of the changes in this PR is good, with two exceptions:
  * The `.project-background` element is all sorts of weird, notably on the home page - a background image shouldn't need its own element nor some complex calculations to display it properly.
  * There's a bit of space after the footer of the project.
* However, as we're going to meet with the BPL on Wednesday, I'm giving this a stamp of "OK for now!".

### Status
Approved and merging, but with a caveat that we need to take another look at our DOM structure (especially in the context of fixing background images) at a later date.

wgranger:
@shaunanoordin Can you create an issue for the background resizing?

PR #24, opened 2017-07-19
'Reset Image' functionality also does auto-best-fit of Image size

### LINKED ISSUE #23: Reset Image should also trigger 'fit Subject image to SVG window' functionality
## Request Overview
* Follows comments in #19 
* Right now, the Reset Image button in ClassifierContainer will reset the transformation of the Subject image, but not call the `SubjectViewer.fitSubjectToContainer()` function that auto-scales the Subject image to fit the SVG element.
  * Ideally, the `resetTransformations()` function in the `subject-viewer` should run the 'auto-scale-to-fit' functionality.
  * Unfortunately, the complication is that we have different bits of data sitting at different places: the SVG element's size is stored in `SubjectViewer` and the image size sits in `SVGImage`. This will need some thinking-majigging.

shaunanoordin:
## PR Overview
* This PR fixes #23 
* With this PR, calling the 'Reset Image' functionality (either on Subject image load, or when clicking the 'Reset Image' button on the Classifier controls) also calls the 'best fit Subject image to viewer' functionality.
* As a result of these changes, the Image Size and Subject Viewer Size data is now stored within the Redux store.

### Status
Ready for review

### DIFF
--- a/src/containers/ClassifierContainer.jsx
+++ b/src/containers/ClassifierContainer.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import {
-  setRotation, setScaling, resetTransformations,
+  setRotation, setScaling, resetView,
   setViewerState, SUBJECTVIEWER_STATE,
 } from '../ducks/subject-viewer';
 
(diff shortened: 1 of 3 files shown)


wgranger:
LGTM. Nice fix.

PR #25, opened 2017-07-20
Fetch Testing Images

wgranger:
Ok, this was pretty straightforward once I got reacquainted with Redux. However, the main question here is, where do we want the call made to Redux to fetch the next subject in a queue.

I also removed all references to Liberating the Liberator, aside from a url in one place as we are still not positive of that address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Liberating the Liberator
+# Anti-Slavery Manuscripts
 
-Liberating the Liberator invites volunteers to transcribe documents from the American abolitionist movement, working together with the Zooniverse community and the Boston Public Library.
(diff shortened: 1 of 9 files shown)


shaunanoordin:
## PR Review
* This PR attempts achieves the following features & changes:
  * The 'Liberating the Liberator' text has been updated to 'Anti-Slavery Manuscripts' across all known pages/components/labels. (NOTE: only the repo name has not yet been changed.)
  * The Subject Redux Duck is now hooked up to project `3012` on `staging`
  * The Subject Viewer component now properly pulls Subject images from a queue, (i.e. from the Panoptes staging API) instead of using a static placeholder image.

So far, this LGTM.

Only one comment for the future:
* We'll eventually need to move the project information (e.g. workflow ID) to a proper dynamic config file.

### Status
Approved, merging.

PR #26, opened 2017-07-20
Write navigator

wgranger:
Write a functioning navigator which mirrors zooming and drag movement on an image.


shaunanoordin:
## PR Overview
* This PR adds a Navigator component to the Classifier page.
* The Navigator component is a miniature "preview" of the Subject image, and mirrors what the Subject Viewer component shows.
* The full feature/functionality list of the Navigator component is as follows:
  * It shows the same Subject image as the Subject Viewer (but way tinier)
  * It shows the "viewport" / "viewing window" of the Subject Viewer as a red rectangle; i.e. when a user is zoomed in on the Subject Viewer, the Navigator gives a general context of which part of the Subject image the user is looking at.
  * By clicking on a specific part of the Subject image on the Navigator, the Subject Viewer should 'jump' its view to that position. **NOT in this PR - this is an intended future feature.**

screen shot 2017-07-21 at 12 24 24
_The Navigator component is to the top-right of the Classifier page. Not how the position and size of red rectangle matches what's being shown on the primary Subject Viewer._

shaunanoordin:
## PR Review
* Currently, there is an issue with matching the position of the red rectangle in the Navigator with the actual view of the Subject Viewer, but only at _really really high zoom levels._
* This is very likely just an issue with the mathematics, e.g. the order of translation first or scaling first. I'll try to take a look into this today, but if not, I'm quite happy to pass this first and then open a separate issue to fix the high-zoom-synchronisation issue.

screen shot 2017-07-21 at 12 24 59
_Note that the Subject Viewer shows the left-most edge of the Subject image, but the Navigator's red rectangle goes beyond the left-most edge. This indicate an excessively large translation() along the x-axis (and a minor drift along the y-axis)._

wgranger:
Let's open an issue about this since we have the basis of the navigator down, then we can tweak the navigator in a future PR.

PR #29, opened 2017-07-21
Add Contrast Button Functionality

wgranger:
This branch allows the contrast button to invert the colors of a subject. I also changed the navigator image to the current subject fetched from the project.


src/containers/ClassifierContainer.jsx
```diff
@@ -114,16 +114,40 @@ class ClassifierContainer extends React.Component {
...
+            </button>
+
+            <button className="flat-button block" onClick={this.useContrast}>
+              <span className="classifier-toolbar__icon">
+                <i className="fa fa-adjust" />
+              </span>
+              <span>Contrast</span>
+            </button>
+
+            <button className="flat-button block" onClick={this.useResetImage}>
```
shaunanoordin: For the buttons that don't do anything _yet_ (Favorite, Collection, SubjectInfo), the `onClick` action should be empty, not `useResetImage`; this temporary functionality doesn't add anything to our testing and is a bit confusing.

src/ducks/subject-viewer.js
```diff
@@ -19,18 +19,20 @@ const MAX_SCALING = 10;
...
-  
+
   //Viewer settings
   viewerState: SUBJECTVIEWER_STATE.NAVIGATING,
   viewerSize: { width: 0, height: 0 },
   imageSize: { width: 0, height: 0 },
 };
 
 //Action Types
+const SET_CONTRAST = 'SET_CONTRAST';
```
shaunanoordin: [minor] Naming suggestion: since the reducer logic is `contrast: !state.contrast`, perhaps a better name for this action is `TOGGLE_CONTRAST`

src/components/SVGImage.jsx
```diff
@@ -59,10 +77,21 @@ export default class SVGImage extends React.Component {
     }
   }
 
+  invertFilter() {
+    if (this.props.contrast) {
+      if (!document.getElementById('svg-invert-filter')) {
+        document.body.insertAdjacentHTML('afterbegin', INVERT);
```
shaunanoordin: Adding the `#svg-invert-filter` SVG definition this way is functional, but it breaks the React pattern. Let me take a look and see if there's a way of doing this in a more React-ish way - maybe inserting the `<svg><defs>` in the SubjectViewer's SVG, or in the Classifier Page HTML or something.

src/containers/ClassifierContainer.jsx
wgranger: Oh yes, this is leftover from my initial intention of adding button functionality to all toolbar buttons.

wgranger:
As always, thanks for the thorough review. I've made the changes suggested. For future note: I originally wrote the invert filter functionality in PFE, and I _believe_ I arrived at the odd solution of using`insertAdjacentHTML()`, etc. to get the filter working in IE. However, it could be that the filter works in IE just fine with this much cleaner solution. If we notice some issues with the filter in IE, we might need to refer back to this PR. 

Currently, I'm unable to check the invert filter on IE because there are some SVG issues on IE and we probably need to make a couple bug fixes to the subject viewer.

shaunanoordin:
👍  Nice work, @wgranger ! This PR looks good to go; I've also made some notes in the IE11 Compatibility issue (issue 18) to note what was done here. There's only one conflict in `ClassifierContainer.jsx` that needs to be resolved, but I'll take care of that.

### Status
Approved, resolving conflicts, and merging.

PR #30, opened 2017-07-21
[For Review] Added a sample Popup

shaunanoordin:
## PR Oveview
* This PR is for @wgranger to review; the important thing is sharing some functional popup modal code, which can then be updated as necessary.
* In this PR, we have an example of how we can create a simple "message popup" for the project.
* The `<Popup>` component is a simple container and can be placed in any container/component you want.
  * Clicking on either the 'close' cutton or on the black background will close the popup.
  * **Of note,** the show/hide logic of the Popup container is actually handled on the parent container/component.
* The style is up for debate and has plenty of room for improvement. E.g. make it fit content instead of trying to fit the screen.
* Possible bug?: the popup _doesn't steal events_ when active, so you can technically still mouse-scroll the background items with the popup open.

### DIFF
--- a/src/components/Popup.jsx
+++ b/src/components/Popup.jsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Popup extends React.Component {
+  constructor(props) {
+    super(props);
+    this.popupBody = null;
+    this.close = this.close.bind(this);
+  }
+  
+  render() {
+    return (
+      <div className="popup" ref={(c)=>{this.popupBody=c}} onClick={(e) => { return e.target === this.popupBody && this.close(e); }}>
+        <div className="popup-title">
+          <button className="fa fa-close" onClick={this.close}></button>
+        </div>
+        <div className="popup-content">
+          {this.props.children}
+        </div>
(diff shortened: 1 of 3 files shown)


src/components/Popup.jsx
```diff
@@ -0,0 +1,46 @@
...
+  close(e) {
+    this.props.onClose && this.props.onClose();
+    return this.stopEvent(e);
+  }
+  
+  stopEvent(e) {
+    e.preventDefault && e.preventDefault();
+    e.stopPropagation && e.stopPropagation();
+    e.returnValue = false;
+    e.cancelBubble = true;
```
wgranger: This is a good opportunity to review these event functions.

wgranger:
I'm going to go ahead and merge this PR, since the main functionality is present and I can begin work on the components of the app that need a popup. I'm also pinging @beckyrother for some design input on the popup.

screen shot 2017-07-24 at 3 27 36 pm

beckyrother:
Thanks @wgranger do you have some time tomorrow to sit down and look at this?

wgranger:
@beckyrother For sure

PR #39, opened 2017-07-27
Functioning Metadata Popup

wgranger:
Closes #35 

This PR adds a simple metadata popup to the classifier. The design is based on the metadata popup in PFE. Hover states are also added to the CSS of toolbar buttons.

### DIFF
--- a/src/components/Popup.jsx
+++ b/src/components/Popup.jsx
@@ -12,7 +12,7 @@ class Popup extends React.Component {
     return (
       <div className="popup" ref={(c)=>{this.popupBody=c}} onClick={(e) => { return e.target === this.popupBody && this.close(e); }}>
         <div className="popup-title">
-          <i className="fa fa-close" onClick={this.close}></i>
+          <button className="fa fa-close" onClick={this.close}></button>
         </div>
         <div className="popup-content">
           {this.props.children}
--- a/src/components/Popup.jsx
+++ b/src/components/Popup.jsx
@@ -7,12 +7,12 @@ class Popup extends React.Component {
     this.popupBody = null;
     this.close = this.close.bind(this);
   }
-  
+
   render() {
     return (
       <div className="popup" ref={(c)=>{this.popupBody=c}} onClick={(e) => { return e.target === this.popupBody && this.close(e); }}>
         <div className="popup-title">
-          <button className="fa fa-close" onClick={this.close}></button>
+          <i className="fa fa-close" onClick={this.close}></i>
         </div>
         <div className="popup-content">
           {this.props.children}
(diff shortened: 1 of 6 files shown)


src/components/Popup.jsx
```diff
@@ -7,25 +7,25 @@ class Popup extends React.Component {
...
     this.close = this.close.bind(this);
   }
-  
+
   render() {
     return (
       <div className="popup" ref={(c)=>{this.popupBody=c}} onClick={(e) => { return e.target === this.popupBody && this.close(e); }}>
         <div className="popup-title">
-          <button className="fa fa-close" onClick={this.close}></button>
+          <i className="fa fa-close" onClick={this.close}></i>
```
shaunanoordin: Query: why are we replacing `<button>` with `<i>` ? With `<button>`, users can use Tab navigation to reach the close button; while `<i>` reduces keyboard accessibility.

However, I can understand this change if the goal is to later add a "Press `Esc` to close popup" feature, which may actually be the best way forward. (Our current keyboard accessibility is sub-optimal, to say the least.)

eatyourgreens: The linter should actually be erroring if `onClick` is added to a text element like `i`, if you're using the same eslint config as PFE.

src/containers/ClassifierContainer.jsx
```diff
@@ -28,10 +29,11 @@ class ClassifierContainer extends React.Component {
     this.useRotate90 = this.useRotate90.bind(this);
     this.useResetImage = this.useResetImage.bind(this);
     this.useContrast = this.useContrast.bind(this);
-  
+    this.showMetadata = this.showMetadata.bind(this);
+
     //TEMPORARY
     this.state = {
-      TEST_POPUP: null,
+      POPUP: null,
```
shaunanoordin: Minor request: (Recommended for code consistency, but not blocking) change the all caps `POPUP` and `CLOSE_POPUP` names to standardCamelCaseNames to better match the rest of the code.

Apologies, I used the all caps naming originally to make sure the methods/variables related to the test popup were super obvious.

src/components/Popup.jsx
eatyourgreens: Seconding @shaunanoordin this element is a button. `<i>` in HTML is used to mark up italic text..

src/containers/ClassifierContainer.jsx
wgranger: Ahh. Ok. Good idea here. I wasn't exactly sure how to proceed if they were temporary, constants, etc. That sounds good though.

src/components/Popup.jsx
wgranger: Yes! Good catch. I was wondering why the linter didn't pick this up. I mainly did this to quickly get rid of some of the pesky default button styling going on, but I can find a workaround to keep the `onClick` handler attached to a button.

eatyourgreens: Maybe set up a global style for `button` that resets it to something that fits the design, then override or add to that for specific buttons.

shaunanoordin:
Nice! Approved, now merging.

PR #42, opened 2017-08-02
Collections Popup

wgranger:
This PR creates the collections manager popup. The form is quite similar to what we have on PFE and also uses react-select. Note: some `!important` styling was necessary to escape default styling set by the component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "react-dom": "^15.4.1",
     "react-redux": "^4.4.6",
     "react-router": "^3.0.0",
+    "react-select": "^1.0.0-rc.5",
     "redux": "^3.6.0",
     "redux-devtools-extension": "^1.0.0",
     "redux-logger": "^2.7.4",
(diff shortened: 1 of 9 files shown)


shaunanoordin:
## PR Review
* This PR allows users to add Subjects to specific Collections.
* This is done by clicking 'Collections' on the Classifier Page toolbar and accessing the Collections popup.
screen shot 2017-08-03 at 13 47 34

While this PR allows performs well under expected success conditions, failure conditions aren't sufficiently accounted for.

### Issues
* [major] A user who is _not logged in_ will...
  * be able to list all available Collections, including ones that they don't own, whaaa?
  * not be able to successful add a Subject to a Collection, but she/he will _not_ receive any indication that any attempt to do so failed.
* [major? medium?] There is no indicator whether adding something to the collection was successful or not - we need to figure out a way to message this.
* [minor] When there are too many Collection names in the dropdown/textbox autofill, it exceeds the size of the popup.

screen shot 2017-08-03 at 13 47 51
_A non-logged-in-user sees a lot of Collections in the dropdown. Also, the size of the dropdown exceeds the size of the popup._

### Status
Changes requested; at minimum:
* Prevent non-logged-in-users from accessing the Collections popup.

I'm OK to leave the other issues for another PR if there's no time to address them.

Issue #28, opened 2017-07-21
Navigator component needs click-to-reposition functionality


shaunanoordin:
## Feature Request
* The Navigator component requires a new feature:
  * By clicking on a specific part of the Subject image on the Navigator, the Subject Viewer should 'jump' its view to that position
* This follows comments on PR #26

wgranger:
Closed with #44

PR #44, opened 2017-08-07
Navigator Rotate and Click Zoom

wgranger:
This PR adds functionality to rotate the image of the navigator along with the subject viewer. It also adds a click zoom to the navigator.


wgranger:
Ah yes. Ok, some good notes above that should be considered. Initially, my construction of this PR was based on #28, but I notice there isn't a zoom functionality there, it was only reposition. Good point about the separation between the subject viewer and navigator.

Concerning the three issues:
1) I could make a change to keep the image stationary at all times and only move the red rectangle when there is a rotation, pan, etc.

2) Based off issue #28, is the idea that a click should reset the image, or center the image to the spot where it was clicked?

3) Yes. I did notice this issue, which seems to be present on master, so a separate PR is probably the best solution here.

shaunanoordin:
Awesomeness 👍  Replying your replies:

1. I think that would be the best - when panning and zooming, keep the image static in the centre of the Navigator, then move the red rectangle around. As for rotation, I'll leave it to you whether you'd like to rotate the image (which may be easier for a human to interpret - check with a UI expert) or the red rectangle (which is way easier to do code-wise).

2. Clicking on the Navigator should centre the image to the spot where it was clicked - basically, just remove that `this.props.dispatch(setScaling(this.props.scaling + ZOOM_STEP));` line of code. 👍 

3. I'll also help look into this once I can free up some time, but that'd be late September at best. Hopefully this issue won't be a breaking matter for our users.

wgranger:
I think the most recent commit covers all the issues we discussed, but I'm open to making any other tweaks, if necessary. Also, I'm curious if issue 3 above is due to some miscalculations on panning the image? It's definitely worth investigating in a future branch.

PR #45, opened 2017-08-07
Add Filmstrip Viewer

wgranger:
This PR adds the filmstrip viewer. Things appear to be working properly, but I'm curious that the svg views are implemented correctly. I'm also curious if my use of frames is correct (reflective of how it is utilized on PFE). Images appear to switch correctly, my only concern is with annotating the correct frame/image, which will probably be tackled in a future PR.

Hardcoded workflow ID was changed to use a subject set with many pages.


src/components/FilmstripViewer.jsx
```diff
@@ -0,0 +1,124 @@
...
+  renderFrames() {
+    const SVG_WIDTH = 40;
+    const SVG_HEIGHT = 60;
+    let scale = 0.08;
+    const viewBox = `-${(SVG_WIDTH / scale) / 2} -${(SVG_HEIGHT / scale) / 2} ${SVG_WIDTH / scale} ${SVG_HEIGHT / scale}`;
+    if (!this.props.currentSubject) { return }
+    const images = getAllLocations(this.props.currentSubject);
+
+    const render = images.map((image, i) => {
+      const activeBorder = i === this.props.frame ? "related-images__frame--active" : ""
```
shaunanoordin: [minor] Missing a `;` here, as well at lines 64 and 69.

src/components/FilmstripViewer.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+      return (
+        <div key={i} className={`related-images__frame ${activeBorder}`} ref={(el) => {this.strip = el; }}>
+          <button onClick={this.changeFrame.bind(this, i)}>
+            <svg
```
shaunanoordin: [Minor?][Medium?] I'm wondering if this `<svg>` element would be as effectively replaced with an `` or a `<div style="background: url(...); background-size: cover;">`

It may be simpler, and we don't have to worry too much about using `let scale=0.08`, which may not be the best fit for all Subject Images.

src/components/FilmstripViewer.jsx
```diff
@@ -0,0 +1,124 @@
...
+    })
+    return (
+      <div className="related-images__frames">
+        {render}
+      </div>
+    )
+  }
+
+  render() {
+    const renderButtons = this.props.currentSubject && this.props.currentSubject.locations.length > 5 ? true : false;
```
shaunanoordin: If we're going to have that `5` there, we might as well declare what it is. i.e. `const MINIMUM_IMAGES_FOR_SCROLL_BUTTONS = 5`

eatyourgreens: The `true : false` is redundant here since that's the value of the conditional on the left of the question mark.

I think you could just assign `renderButtons = this.props.currentSubject && this.props.currentSubject.locations.length > 5`

src/components/FilmstripViewer.jsx
simensta: If you don't need the SVG capabilities, I would replace the `<svg />` element with an `` element. In either case, can you add an alt attribute to describe the image frame?

src/components/FilmstripViewer.jsx
wgranger: Nice. Still need to figure out what's going on with the linter.

shaunanoordin:
## PR Review (Update)
* The latest commits address all the minor issues from the last PR Review, and as such this PR is ready (readier?) to merge.
* HOWEVER, I've discovered a new issue with the film strip viewer implementation: the filmstrip viewer demands ALL hi-res Subject Images on load, causing significant loading times.
  * HOWEVER HOWEVER, this is a separate issue to be investigated; I'll write something up separate from this PR.
    * HOWEVER however HOWEVER, that issue actually already has a solution thanks to Cam, so it's probably less of a problem to solve, and more of just another thing on the TODO list.

### Status
Approved & merging. Please keep an eye out for the separate Filmstrip Viewer Load Times issue.

PR #46, opened 2017-08-07
Fix Errors with addToCollection

wgranger:
This takes care the issue of the collection popup closing before displaying errors with collection adds.


shaunanoordin:
## PR Review
* This PR fixes the the issue where users would not be aware if there were any problems when adding a Subject to a Collection.
screen shot 2017-08-09 at 14 56 49
_Error shown when attempting to add the same Subject to a Collection that the Subject is already a part of._

### Status
LGTM! Merging

PR #51, opened 2017-08-14
Use Thumbnail Sizing

### LINKED ISSUE #48: Filmstrip Viewer Load Time
## Fix Required
* This follows PR #45
* With the current implementation of the Filmstrip Viewer and multi-image Subjects, the Filmstrip viewer requests _all the images_ from the Subject _on load._
  * This means that as soon as you open the Classifier Page, you might be hit with, say, 10x requests for 1.2MB image - the page will load 10.2 freaking MB before you do anything.
  * This is bad for the users for a multitude of reasons, but the most pertinent to the project is that (if the initial Subject image isn't the first to be fully loaded) it could take users a whole minute before they can even start annotating.
* Fortunately, there is a solution on the horizon:
  * Use thumbnails for the Filmstrip Viewer instead of the full-res Subject images.

### Recommended Solution
* As Cam pointed out, the Zooniverse already has a thumbnail generatator-majig. (nginx?)
* For example,
  * If we want to convert the Subject image https://panoptes-uploads.zooniverse.org/staging/subject_location/87e48d7c-0a30-4ec7-9d31-47361a68db8f.png into a thumbnail,
  * we use this URL: https://thumbnails.zooniverse.org/250x250/panoptes-uploads.zooniverse.org/staging/subject_location/87e48d7c-0a30-4ec7-9d31-47361a68db8f.png
* This thumbnail URL would replace the full-size image URLs used in the Filmstrip Viewer component.

Further documentation, notably for setting the width/height/etc: http://nginx.org/en/docs/http/ngx_http_image_filter_module.html

### Further improvements 
* [minor] Currently the filmstrip viewer uses a hardcoded inline style to set the image to 40x60px (  `` )
  * This can be replaced with `width: 100%; height: auto;` to prevent any squishing/stretching of the preview image if that image doesn't match the 2:3 size ratio.

Tagging @wgranger ; while not a core/critical issue, addressing this will be a priority (categories: performance, usability) closer to launch or when the project has more testers/users on it.

wgranger:
Closes #48 

This PR uses thumbnail versions of an image to cut back on load time of the FIlmstrip Viewer when opening the classifier.

### DIFF
--- a/src/components/FilmstripViewer.jsx
+++ b/src/components/FilmstripViewer.jsx
@@ -5,6 +5,9 @@ import { connect } from 'react-redux';
 import { getAllLocations } from '../lib/get-subject-location';
 import { changeFrame } from '../ducks/subject-viewer';
 
+const THUMB_WIDTH = 40;
+const THUMB_HEIGHT = 60;
+
 class FilmstripViewer extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


shaunanoordin:
## PR Review
* This PR replaces the full-size images used in the Filmstrip Viewer with smaller thumbnail images.
* This solves issues as listed in #48 , notably improving load times.

### Status
LGTM! Merging.

PR #53, opened 2017-08-17
Move Zoom Tools to Subject Viewer

### LINKED ISSUE #22: Add proper mouse cursors for Subject Viewer
## Request Overview
* In the Subject Viewer, when a user is navigating the Subject (via mouse click-drag), the mouse cursor should transform into something more informative.

### LINKED ISSUE #52: Move Zoom buttons
Zoom in/out and drag hand buttons should be moved over to the top left size of the subject viewer and removed from the toolbar.

01-transcribe

wgranger:
Fixes #52 Fixes #22 

Per design suggestions, this PR moves zoom and pan functionality to the subject viewer. The cursor has also changed to a move icon when navigating.

### DIFF
--- a/src/components/ZoomTools.jsx
+++ b/src/components/ZoomTools.jsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { SUBJECTVIEWER_STATE } from '../ducks/subject-viewer';
+
+const ZoomTools = ({ viewerState, usePanTool, useZoomIn, useZoomOut }) => {
+  return (
+    <div className="zoom-tools">
+      <button className="flat-button block" onClick={useZoomIn}>
+        <i className="fa fa-plus" />
+      </button>
+
+      <button className="flat-button block" onClick={useZoomOut}>
+        <i className="fa fa-minus" />
+      </button>
+
+      <button
+        className={(viewerState === SUBJECTVIEWER_STATE.NAVIGATING) ? 'flat-button block selected' : 'flat-button block'}
+        onClick={usePanTool}
+      >
+        <i className="fa fa-arrows" />
+      </button>
+    </div>
+
+  );
+};
+
+ZoomTools.propTypes = {
+  usePanTool: PropTypes.func,
+  useZoomIn: PropTypes.func,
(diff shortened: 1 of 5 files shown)


shaunanoordin:
## PR Review
* This PR is functionally a re-arrangement (and minor improvements) of a few UI elements.
* The Zoom In/Out buttons, as well as the "Pan/Move" button, has been moved to the left side of the screen.
* The Subject Viewer now correctly uses a 'Move' mouse cursor when in Pan/Move mode.
screen shot 2017-08-21 at 12 16 18

The code LGTM, but I do have some UI/UX thoughts. (Edit: why can't I ping Becky in on this? I can't find her name on the @ list)

### UI Comments
* The Classifier Page has two distinct 'modes': Viewing (panning/moving) and Annotating.
* However, the integration between the two and the visual signalling between them seem a bit disjointed.
  * _We'll ignore the issue of integration for the moment (i.e. making Viewing and Annotating very simple/intuitive to switch between) as that's another ten paragraphs of thoughts_
* As for the visual signalling, what we need to indicate is that there's a 'switch' (in the light switch sense) for users to click to switch between Viewing and Annotation.
* This 'switch' is separated into two buttons, the Annotation button and the Pan/Move button.
* However, with this PR, those two buttons are separated spatially on the visuals.
* This is akin to having a button to switch off the living room lights at the front of the room, but to switch them back on, you'll need to cross the room to press a different button at the back of the room.

Strictly speaking, this doesn't break functionality, but there's a part of me that says we'll need to revisit this design implementation again soon.

### Status
Code looks good, so I'm happy to merge this. However, the UI will probably need to have a few further iterations in the future for improvement.

beckyrother:
@shaunanoordin thanks, this makes sense! 

I moved the pan tool near the zoom tools because those tools all have to do with the **view** of the subject. The Add Annotation tool and the rest of the toolbar has to do with **editing** the subject. So rather than thinking of it as a Switch, separating the two tools makes it very clear to the user that they need to use the Annotate tool to actually add an annotation. 

Is the hesitation to do with the placement of the zoom and pan tools? Would it make more sense to move them to the upper right of the screen, closer to the rest of the controls?

PR #56, opened 2017-08-24
Style Annotations Dialog

### LINKED ISSUE #40: Build a proper Annotation Text Editor popup/dialog box/etc
## Feature Needed
* This follows the Annotation Interface PR at #32 
* The Annotation Text Editor in that PR is only a placeholder, showing how an Annotation can be selected using either `ducks/annotations.selectAnnotation()` or `ducks/annotations.completeAnnotation()`
* A proper Annotation Text Editor needs the following features.
  * As a user, I have a textbox in the ATE to transcribe a line of text.
  * As a user, I can click the 'Save' button close the ATE and commit the changes.
    * This will likely require a new function such as `ducks/annotations.updateTextOfSelectedAnnotation(newText)`
  * As a user, I can click the 'Close/Cancel' button to close the ATE without committing changes.
  * As a user, I can click the 'Delete' button to close the ATE and delete the currently viewed annotation.
    * Possible new function: `ducks/annotations.deletedSelectedAnnotation()`
  * As a user, while the ATE is open, I must be able to see the Annotation marks/the _visual line of text_ that I just marked.
    * There are two ways to do this:
      * One, the ATE is a repositionable dialog box within the Classifier page (either it can't be a <Popup> component, or the <Popup> component needs to be modified to be repositionable).
      * [advanced] Two, the ATE has an SVG mini-window previewing that line of text.

wgranger:
Closes #40 

I believed I resolved the issue with placement of the annotations pane on a selected annotation. The initial issue is that the absolute placement values of the pane was being set awkwardly according to the parent, so moving the pane to the subject viewer helped make things more approximate. 

There is currently an issue of the pane displaying off screen, but that can be solved in a future PR. Although the pane placement could have been solved, this might still be a WIP in case there is something further I am missing on the annotations pane.


shaunanoordin:
Also, are we OK to remove that [WIP?] tag from the title?

wgranger:
I think we can remove the WIP tag. I feel more confident about the dialog placement after the most recent commit. My main priority is to get this merged today or tomorrow (at the latest) so I can begin to work on annotations.

wgranger:
The issue with typing in the input was that the parent component `<Rnd />` had a drag handler onClick which was affecting each child.

shaunanoordin:
@wgranger Nice catch with the Rnd drag handler! 👍 

The text input works - I can type into the <input /> text box properly now. That's good!

However, clicking on the "Done" button does nothing. That's bad!

However, clicking on the "close" button actually saves the text you typed. That's good!

However, clicking on the "Cancel" button does the same thing as the "close" button, so you're not cancelling your input, but committing it. That's bad!

https://youtu.be/Krbl911ZPBA?t=22s

shaunanoordin:
Only one Change Request to address to get this approved:
* For the time being, just remove the non-functional "Done" button, then rename "Cancel" to "Done". Problem solved for now - we can revisit proper Cancel button functionality later.

wgranger:
Yes, button functionality was what I was aiming to work on this week with saving annotations and classifications.

shaunanoordin:
Wait what? Conflicts? Let me see...

shaunanoordin:
Re: **package.json**,
* `"react-rnd": "^5.1.0",` seems to be the correct version.

Re: ** ClassifierContainer.jsx**,
* Only `import { toggleDialog } from '../ducks/dialog';` needs to be kept.
* `import { ... } from '../ducks/annotations';` isn't needed for the container.

Re: **popup.styl**,
* I have no idea, unfortunately.

wgranger:
Should be all set

PR #57, opened 2017-08-29
Restyle Metadata Popup and Create Dialog

wgranger:
Some styling changes were made to the metadata popup. Also, the popup itself was changed from a modal into a popup that could be moved around. This was also done by moving the popup into redux and making it available on the root page so absolute positioning would place the popup consistently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "react": "^15.4.1",
     "react-dom": "^15.4.1",
     "react-redux": "^4.4.6",
+    "react-rnd": "^5.1.0",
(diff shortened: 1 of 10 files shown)


shaunanoordin:
I just realised - is `src/components/App.jsx` missing its initial props?

wgranger:
Good call on the white background. There is actually a drop shadow in the mockup, so I'll add that in. I think the second issue (marking new annotations with a box still open) is worth opening an issue, as this will take a bit of thinking through.

PR #58, opened 2017-08-29
Toggle Previous Marks Functionality

wgranger:
The ability to toggle previous marks was added to the classifier and works through Redux.


wgranger:
The Toggle Prev. Marks button should remove marks by all users, both of the current user and all other users.

PR #60, opened 2017-08-30
Fix IE Issues

wgranger:
IE has some knows issues with Flexbox, so many of the changes here were in response to those issues. Also, IE doesn't support `.include()` when fetching an image src, so `.indexOf()` was used in place.

### DIFF
--- a/src/components/SocialSection.jsx
+++ b/src/components/SocialSection.jsx
@@ -4,7 +4,7 @@ import Divider from '../images/img_divider.png';
 
 const SocialSection = ({ classifications, completedSubjects, percentComplete, subjects, volunteers }) =>
   <div className="home-page__social-section">
-    <h2 className="main-title">Subjects <i className="secondary-conjunctions">of</i>  Note</h2>
+    <h2 className="main-title">Subjects<i>of</i>  Note</h2>
     <img role="presentation" className="divider" src={Divider} />
     <div className="home-page__subject-row flex-row">
       <div></div>
--- a/src/index.tpl.html
+++ b/src/index.tpl.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="x-ua-compatible" content="ie=edge">
     <title>Anti-Slavery Manuscripts</title>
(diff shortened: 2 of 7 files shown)


srallen:
If you want to use newer spec like Array.includes(), there're lots of polyfills out there and cdn services. We use polyfill.io's cdn for polyfilling on PFE: https://polyfill.io/v2/docs/

shaunanoordin:
I agree with Sarah; for functions, using polyfills is the way to go. 👍 

Actually, I think we already have a polyfill? https://github.com/zooniverse/liberating-the-liberator/blob/master/src/index.tpl.html#L7 But I never checked if it was working properly for all the features we need, so polyfill.io is probably the more reliable way to go.

The Flexbox issues, however, are a different story - let me take a look.

wgranger:
Interesting. I'm not familiar with using one of these polyfill services, but the docs look thorough, so I can look into using an `.includes()` through a polyfill if that's a better option.

srallen:
polyfill.io's defaults cover a lot, but some have to specified like Array.prototype.includes. Check out what we're doing with PFE: https://github.com/zooniverse/Panoptes-Front-End/blob/master/views/index.ejs#L73

wgranger:
Well if that wasn't just the easiest thing I've ever done.

shaunanoordin:
## PR Review
* This PR fixes compatibility issues with IE11 and MS Edge.
* Accessing the localhost on Win7+IE11 and Win10+Edge, we see that...
  * The main pages - Home, Classifier, About the Collection - look visually OK on a _desktop_ resolution. There are no visible issues with the layout or the like.
  * Basic interactions with the Classifier page have also tested OK, notably: adding Annotation.

Overall, this LGTM.

Side discussion:
@srallen You're right, it _is_ the same service (visiting the root Heroku domain shows the Polyfill.io website), and since I was the one who added the link (according to Blame), it means I somehow stumbled across the wrong URL for the correct service. To explain this discrepancy, I'm just going to say... a wizard did it.

### Status
LGTM, approving & merging. And remember, if this PR inexplicably causes problems in the future: https://youtu.be/sVgVB3qsySQ?t=16s

PR #61, opened 2017-09-05
Save classifications

wgranger:
(9/5/17) Currently, this is WIP for a few reasons. Right now, the PR is capable of saving a classification in the Panoptes backend. The lines below may be enough for a merge (with a bit of cleaning up) unless I'm missing an integral part of how classifications are saved. At this point, I'm mainly looking for feedback on this branch.

- Does our metadata follow a required structure (recording image width/height, etc.)
- TODO (before/after?): the `get-session-id` functions need to be fleshed out.
- That said, there needs to be some structure of saving annotations and coming back to a session later (the following day, etc.) if a classification is not complete.


package.json
```diff
@@ -24,6 +24,7 @@
   },
   "homepage": "https://github.com/zooniverse/liberating-the-liberator",
   "dependencies": {
+    "counterpart": "^0.18.1",
```
shaunanoordin: What's this doing here? Isn't this a library for handling multiple languages in our UI?

shaunanoordin:
While the code looks fine, I can't comment much about the functionality until I can find a way to test/confirm Classification submissions. I'll probably need to fiddle around with the ZooAPI to 'fetch all Classifications' or something.

Side note: it may help if we actually document the output (Classification) data format we want sent out, since - given how fluid the details are - we might forget and then accidentally look back in a month and go, "oh crap, we were supposed to have a bit of metadata for each line of Annotation indicating whether the user looked at the Aggregations or not."

package.json
wgranger: I included this because we're saving language information in the metadata of a classification with `counterpart.getLocale()`. With this PR, I am constructing metadata similar to how it is set on PFE.

wgranger:
We should probably have a conversation about metadata. This was one of the main unknowns when I opened this PR because I was a bit unsure if we wanted the same info as in PFE or if there was some additional info we wanted to record.

wgranger:
Note: I'm taking off WIP although there is still a bit of work that still needs to be done towards classifications. Namely, metadata needs to be nailed down, but this shouldn't be a blocker to saving classifications.

Also, this branch does not save a new classification when a previous annotation is selected. I refrained from adding that functionality until we get whole line results from the extracts instead of single words.

srallen:
Just from past experience by not following PFE classification convention with the initial Wildcam Gorongosa build, I would suggest following it as closely if not exactly for the classifications. Panoptes will not validate what's in the metadata or annotations, but auxiliary services like aggregation and caesar may depend on convention, so we should follow it for custom front ends too.

wgranger:
Thanks for the input. I think we should be good for this PR, and I should probably clarify further what I did with classifications in this example.

* In this PR, points are saved exactly as the ellipse tool saves coordinates on PFE.
* The metadata field only has a couple tags that are missing from PFE examples. However, Caesar appears to be working well and raw classifications are returned from data exports. 

That said, it's important to note that the test workflow used in this repo also has an ellipse tool defined so exports will expect the correct tool points in the Panoptes backend.

src/ducks/annotations.js
```diff
@@ -118,11 +134,11 @@ const selectPreviousAnnotation = (data) => {
   };
 };
 
-const addAnnotationPoint = (x, y) => {
-  return (dispatch) => {
+const addAnnotationPoint = (x, y, frame) => {
+  return (dispatch, getState) => {
```
shaunanoordin: [minor] getState isn't needed here, is it?

src/ducks/classifications.js
```diff
@@ -0,0 +1,91 @@
...
+      return state;
+  }
+};
+
+const createClassification = () => {
+  return (dispatch, getState) => {
+    const classification = apiClient.type('classifications').create({
+      annotations: [],
+      metadata: {
+        workflow_version: getState().workflow.data.version,
```
shaunanoordin: [major] Problem: the default value for `workflow.data` is `null`, so certain calls for `workflow.data.version` will fail. Recommended solution: check for the existence of `workflow.data` and provide a default value if that fails.

src/ducks/classifications.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+  };
+};
+
+const submitClassification = () => {
+  return (dispatch, getState) => {
+    const annotations = {
+      _key: Math.random(),
+      _toolIndex: 0,
+      task: getState().workflow.data.first_task,
```
shaunanoordin: [major] Potential problem: this line runs the risk of `workflow.data` being null when `.first_task` is called. I'd also recommend checking the existence of `data` first, and providing a default value otherwise.

src/ducks/classifications.js
```diff
@@ -0,0 +1,91 @@
...
+      value: getState().annotations.annotations,
+    }
+
+    dispatch({
+      annotations,
+      type: SUBMIT_CLASSIFICATION,
+      task: getState().workflow.data.first_task
```
shaunanoordin: [major] Potential problem: this line runs the risk of `workflow.data` being null when `.first_task` is called. Same as above.

src/lib/get-session-id.js
```diff
@@ -0,0 +1,45 @@
+// const stored = sessionStorage ? sessionStorage.getItem('session_id') : null;
```
shaunanoordin: Since this whole file is commented out, is this file a placeholder?

shaunanoordin:
I'm getting several errors in regards to the workflow data, e.g. the Classification Page can't tell what's the Workflow version. I'm wondering if this is because I'm not logged in, and the Workflow isn't accessible to the public.

shaunanoordin:
## PR Review
* This PR adds the ability for the user to **submit a Classification** (i.e. a collection of lines-of-text Annotations for a specific Subject) to Panoptes via the Classification page.
* This is done by creating some Annotations on the Classifications page, then clicking the 'Done' button.
* Notes:
  * For testing purposes, the fix in #73 has been temporarily applied when testing this PR locally. (This is to save time from rebasing this entire branch.)
  * This PR has an external dependency: a functioning connection to Panoptes.

Unfortunately, while testing on localhost+OSX+Chrome60, I encountered some notable issues.

### Issues

**Issue 1:** "cannot read property 'version' of null" / Classification Ducks don't check for variable existence.
screen shot 2017-09-26 at 15 44 07

Analysis:
* In `ducks/classifications.js`, there are numerous instances of variables being called without checking for their existence.
* For example, in the line `workflow_version: getState().workflow.data.version,`, there are no safety checks to ensure that `workflow.data` _isn't_ null before calling `.version`. (The default value for workflow.data is null.)

Recommended Solution:
* Run the usual existence checks, and provide a default value otherwise.

This might be related to...

**Issue 2:** Cannot submit Classifications / Workflow data doesn't exist
* When clicking 'Done' to submit my Classifications, Panoptes returns a 422 Unprocessable entity error.
screen shot 2017-09-26 at 16 39 33

Analysis:
* Clearly, this stems from the fact that the app has no idea what Workflow version needs to be submitted to Panoptes.
screen shot 2017-09-26 at 16 40 47
_Example of the data submitted to Panoptes. Note that the Workflow version is null._

While this existence checks definitely need to be in place, the Workflow data _should_ have been retrieved. My suspicion is that the Workflow isn't accessible to the public.

Recommendation:
* Investigation required. Attempt to submit Classifications while in an incognito window, not signed in, to examine theory that the Workflow isn't accessible to the public, then contrast it by submitting while logged in as @wgranger or other owner/collaborator of the Zooniverse Project.

### Status
This PR requires that the two major issues be addressed before it can be approved.

Next step of the review, once issues are addressed, would be to confirm that the submitted data follows the necessary data format and contains all the information required by the research team _and_ the aggregation systems.

src/lib/get-session-id.js
wgranger: Yes, I left it as a placeholder because I felt this work was most appropriately done for a future PR rather than making this PR too fat.

wgranger:
Ok, this PR has been updated recently. As far as I know, the only two issues remaining is saving `subject_dimensions` in the metadata. I'll direct you to how it's done in in PFE with the [classification creation](https://github.com/zooniverse/Panoptes-Front-End/blob/6a228205f1aa258c76fb0d5e79459c246f6abb24/app/pages/project/classify.cjsx#L148) and [dimension updates](https://github.com/zooniverse/Panoptes-Front-End/blob/0f3b9c02c4b8e031bc5e4fa8d3674ea170c9281f/app/classifier/classifier.jsx#L146).

shaunanoordin:
## PR Update
* This PR has now been updated so that it records `naturalWidth` and `naturalHeight` metadata (i.e. the actual size of the original image).
* It does NOT, however, provide an accurate record of `clientWidth` and `clientHeight` (i.e. the visible size of the image on the user's web browser) due to implementation concerns.

### Dev Talk
@wgranger , after trying to record `clientWidth` and `clientHeight`, I ran into a few stumbling blocks, namely: 1. getting the clientWidth/height of a React-SVG <image> without resorting to DOM hacks, and 2. finding the best places to trigger `setImageMetadata()` for the client sizes. (When the zoom changes? On the first resetView?)

While solvable, I'd rather spin the clientWidth/clientHeight record into a later PR, after Sam has a chance to speak with the BPL researchers so I don't spend more time on a feature we're not sure they want.

### Original Discussion
From the #imls channel on Slack:
> Question: How important is it for the BPL researchers to know the metadata of the actual image size vs the size the user sees it as on his/her web browser?
> Context: Standard zooniverse.org projects record `metadata.naturalWidth`, `metadata.naturalHeight`, `metadata.clientWidth`, `metadata.clientHeight`. (natural size = actual size of image. client size = the size of the image as shown on the user's browser. So if the user is viewing a large image on a tiny handphone, for example: naturalWidth=2000, clientWidth=800)
> Standard zooniverse.org projects record this, ostensibly, because researchers want to understand _user behaviour_ - if the user is looking at a scaled down image, is she/he missing out on a crapton of detail?
> In the context of ASM, this kind of _recording of user behaviour_ doesn't work as well with such simple metadata, simply because ASM has _more complicated methods of interacting with a Subject image._ Namely, zoom/rotate/translate - the standard clientHeight/clientWidth metadata (the size of an image as it appears on the web browser) doesn't make sense because with ASM, said image can be super zoomed in for one Annotation, then rotated for a second Annotation, zoomed out for a third, etc.

### Status
Ready for... re-review? It's a minor update on top of a stable PR, so normally I'm happy to merge it straight away, but it's probably good to have a quick check on the last two commits, to make sure I didn't simply give myself A+ for my own "minor updates". (What _is_ the review etiquette for reviewing something you did that's based on something else someone else did that they asked you to review?) 😵

src/ducks/classifications.js
```diff
@@ -37,20 +37,23 @@ const classificationReducer = (state = initialState, action) => {
...
   return (dispatch, getState) => {
-    let version = "";
+    let workflow_version = '';
     if (getState().workflow.data) {
-      version = getState().workflow.data.version;
+      workflow_version = getState().workflow.data.version;
     }
-    const dimensions = [];
+    
+    let subject = getState().subject;
```
wgranger: Maybe I'm missing something. It doesn't look like this variable is used anywhere below.

shaunanoordin: Thanks for catching this - the error is larger than it appears. I seem to have lost the changes I made for classifications.js in a stash/pop mishap, so I'll need to put back in the code that was lost. Updates incoming.

shaunanoordin:
## PR Update
* A new fix has been added to _actually submit the image size metadata_ along with the Classification.
* Note: I overlooked stating this before, but while the Submit Classifications mechanisms are in place, moving to the next Subject to Classify (i.e. the post-submit cleanup) is still on the ToDo list.

### Request Payload Sample
```
{  
  "http_cache":true,
  "classifications":{  
    "annotations":[  
      {  
        "task":"T1",
        "value":[  
          {  
            "details":[  
              {  
                "value":"No 7 B Oakes"
              }
            ],
            "points":[  
              {  
                "x":587.8937097385373,
                "y":257.8584691170898
              },
              {  
                "x":669.6746873815839,
                "y":263.43444486547935
              },
              {  
                "x":731.0104206138689,
                "y":261.57578628268277
              },
              {  
                "x":937.3215233042819,
                "y":270.8690791966654
              }
            ],
            "frame":0
          },
          {  
            "details":[  
              {  
                "value":"Absolutely_nothing"
              }
            ],
            "points":[  
              {  
                "x":1343.4551104972377,
                "y":460.0587016574585
              },
              {  
                "x":916.6388121546961,
                "y":458.70372928176795
              }
            ],
            "frame":1
          },
          {  
            "details":[  
              {  
                "value":"some slick man"
              }
            ],
            "points":[  
              {  
                "x":812.3059392265193,
                "y":395.02002762430936
              },
              {  
                "x":930.1885359116022,
                "y":388.2451657458563
              },
              {  
                "x":1107.6899171270718,
                "y":390.95511049723757
              },
              {  
                "x":1193.0531767955802,
                "y":392.31008287292815
              }
            ],
            "frame":0
          }
        ]
      }
    ],
    "metadata":{  
      "workflow_version":"10.13",
      "started_at":"2017-10-09T16:18:57.288Z",
      "user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36",
      "user_language":"en",
      "utc_offset":"-3600",
      "subject_dimensions":[  
        {  
          "naturalWidth":1962,
          "naturalHeight":1038,
          "clientWidth":1962,
          "clientHeight":1038
        },
        {  
          "naturalWidth":1962,
          "naturalHeight":1038,
          "clientWidth":1962,
          "clientHeight":1038
        }
      ],
      "finished_at":"2017-10-09T16:20:13.946Z",
      "viewport":{  
        "width":1064,
        "height":921
      }
    },
    "links":{  
      "project":"1764",
      "workflow":"3017",
      "subjects":[  
        "72815"
      ]
    },
    "completed":true
  }
}
```

screen shot 2017-10-09 at 17 27 05
_Sending the Classification to Panoptes Staging results in a 201 response_

### Status
Thanks for the catch @wgranger , this is now re-ready for re-review.

shaunanoordin:
Update: resolved some merge conflicts from the recent merging of PR #78 and #80

wgranger:
Looks good

Issue #64, opened 2017-09-08
Determine Dot Colors When Selected, etc.


wgranger:
@beckyrother

This is a minor UI need. What color will distinguish the following annotation dots:

- Previous clustered annotations
- User's individual annotations
- User's selected annotations

The image below is used for reference.

![screen shot 2017-09-08 at 10 47 37 am](https://user-images.githubusercontent.com/14099077/30219930-3f530792-9483-11e7-9c22-c043de54f39e.png)

beckyrother:
I'm happy with these colors, as well as the blue you're using for your own previous annotations (see image).
![anti-slavery_manuscripts](https://user-images.githubusercontent.com/1392130/30562109-f438ea8a-9c82-11e7-816c-f1d8e3202920.png)

beckyrother:
Are there other colors I'm forgetting?

wgranger:
To make sure we've got this nailed out:

Your current and previous annotations: cyan
Selected clustered annotation: green
Unselected clustered annotation: red

Is there a difference between clustered annotations that are retired and not? Line retirement seems a bit iffy, but I know we talked about it. 

I say "iffy" because we would just be changing the color of a line depending on how many people had annotated and maybe disabling the dropdown from displaying on a click. A user would still be able to add an annotation on that line by making new marks if (s)he wants to.

wgranger:
Although some work was done in #81, we still need to add greyed out lines for finished rows.

PR #66, opened 2017-09-08
Retrieve and Display Annotations from Caesar

wgranger:
Things seem to be working well on this branch, however, some changes will need to be made in the future depending on how Coleman's code displays line clustering rather than single points. 

Also, we did not discuss structure of this code too much, so I'm open to ideas. Some of the code here operates based on what is included in #61


src/components/AnnotationsPane.jsx
```diff
@@ -71,7 +73,7 @@ class AnnotationsPane extends React.Component {
         svgPoints.push(
           <circle
             key={svgPointPrefix+i}
-            cx={point.x} cy={point.y} r={10} fill="#3fc"
+            cx={point.x} cy={point.y} r={10} fill="#5cb85c"
```
wgranger: I went ahead and made these changes, but it really depends on how #64 is resolved.

src/ducks/aggregations.js
```diff
@@ -0,0 +1,49 @@
+import { request, GraphQLClient } from 'graphql-request';
+
+const initialState = {
+ data: null,
+ selectedPreviousAnnotation: null
+};
+
+const TEMP_WORKFLOW_ID = '3017';
+const TEMP_SUBJECT_ID = '72815';
+const CAESAR_HOST = 'https://caesar-staging.zooniverse.org/graphql';
```
shaunanoordin: These temporary constants are functional for testing, but we may need to start setting up our dynamic config file with these values [(example config.js)](https://github.com/zooniverse/edu-api-front-end/blob/master/src/lib/config.js) soon before we forget.

src/ducks/subject.js
```diff
@@ -14,7 +14,7 @@ const FETCH_SUBJECT_SUCCESS = 'FETCH_SUBJECT_SUCCESS';
 const FETCH_SUBJECT_ERROR = 'FETCH_SUBJECT_ERROR';
 const TOGGLE_FAVORITE = 'TOGGLE_FAVORITE';
 
-const TEMPORARY_HARDCODED_WORKFLOW_ID = '3022';
+const TEMPORARY_HARDCODED_WORKFLOW_ID = '3017';
```
shaunanoordin: As noted in another comment, this hardcoded value needs to be moved to a config file soon-ish, before we forget.

PR #73, opened 2017-09-25
Fix 72: AnnotationsPane now checks for currentAnnotations.clusters_text

### LINKED ISSUE #72: Classifier Page (AnnotationsPage) has rendering errors
## Issue
* When accessing `localhost:3000/classify`, the console occasionally spews out the _"you're trying to get the property of an undefined variable"_ error.

Steps:
* Go to `localhost:3000`.
* Click on the Transcribe link.
* See the following error in the console: `Uncaught TypeError: Cannot read property 'map' of undefined (AnnotationsPane.jsx?aade:173)`

This gets worse when you travel from the Transcribe page, to another page, then back to the Transcribe page again - the error means you're unable to access the Transcribe page twice in one session. (Reload the page to reset.)

screen shot 2017-09-25 at 17 34 21
_The error in question. Note that the Subject photo struggles to be loaded_

Given the description, yeah, I'm going to assume that there's a `something.map()` call somewhere that accidentally skipped the `!something` check. Should be an easy fix, brb.

### Status
@shaunanoordin is investigating

shaunanoordin:
## PR Overview
* Fixes #72 
* Classifications page on localhost:3000/classify would have issues rendering some Subjects (presumably, those without previous markings)
* Turns out this was due to `AnnotationsPane.renderPreviousAnnotations` not checking the existence of `currentAnnotations.clusters_text` before trying to `.map()` it.
* Fix is a simple check before execution.

### NOTE
* @wgranger it occurs to me that the reason I didn't catch this earlier was because the `currentAnnotations.clusters_text` was _always_ provided... which hints that Caesar is likely now _live_ with the new text aggregation updates, wooo ( https://github.com/zooniverse/aggregation-for-caesar/pull/34 )

### Status
Ready for review just in case there's anything else obvious I missed. Otherwise, since this is minor, I'll merge within 24 hours.

### DIFF
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -169,7 +169,7 @@ class AnnotationsPane extends React.Component {
       const annotationData = [];
       const currentAnnotations = reduction.data[`frame${this.props.frame}`];
 
-      if (currentAnnotations) {
+      if (currentAnnotations && currentAnnotations.clusters_text) {
         currentAnnotations.clusters_text.map((text, i) => {
           const data = {
             points: [{


shaunanoordin:
EDIT: just checked Slack, and I realise now that @wgranger is already aware of this issue. (Sorry, I missed Friday's updates.) This PR is now marked irrelevant and closed, as I assume Will is already working on a more generalised PR that will update the Annotations Pane to handle the new data format.

However, in addition to said data format updates, please remember to also run the necessary existential checks before running, e.g., `.map()` functions. Simultaneously, I'll make a note to double-check for this when I run my PR reviews next time.

shaunanoordin:
This PR has been resurrected and will be merged immediately.

I just had a talk with Will, and while he is working on a new branch with the relevant Caesar JSON data format updates, that will take a while and the data format is _still expected to change further._ This means we need a fix for the front-end crashing ASAP, so this should take care of that.

Also, to reiterate what I mentioned on Slack: the front end should _never_ crash if an external service is unavailable. It should either fall back to a failsafe (e.g. if Caesar is down, we should still be able to do normal Classifications... and the metadata should record that the aggregation service was unavailable) or fail gracefully (e.g. if Panoptes is down, we need a "lol sorry, no Zooniverse for you" message)

PR #78, opened 2017-10-05
Prepare Splits for Visibility Tests

wgranger:
This PR sets up retrieval of seven-ten split data in preparation for visibility splits. Some more visual additions need to be merged into this repo before visibility components can be added.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { ZooFooter } from 'zooniverse-react-components';
 import { fetchProject } from '../ducks/project';
+import { fetchSplit } from '../ducks/splits';
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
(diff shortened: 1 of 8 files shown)


src/components/App.jsx
```diff
@@ -16,6 +17,12 @@ class App extends React.Component {
     this.props.dispatch(fetchProject());
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.user != nextProps.user) {
```
shaunanoordin: [minor?] This will trigger when the user has logged in, (`user` goes from `null` to an object) which is expected, but it'll also trigger when the user logs out, (`user` goes from an object to `null`) correct? Is this intended?

This is a minor comment at the moment because I've testing logging in/out and the app works fine. I do see a `401` error from seven-ten when I logout, but so far this hasn't caused any issues on our app itself. If seven-ten _does_ have an issue with asking for splits for a null user, however, this should be easily rectified by adding `...&& nextProps.user !== null` to that `if` clause.

PR #81, opened 2017-10-09
Update Caesar Data

wgranger:
This code is an update of the data that arrives from Caesar. A few helper methods are added to help with the construction of sentences and annotations from previous annotations.

After this, the next step will be to have annotations save from a previous annotation by clicking "done" on the annotations pane.


src/lib/construct-line.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+function constructLines(line) {
+  const lines = [];
+  line.clusters_x.map((point, i) => {
```
shaunanoordin: Do we need to account for `line` being null? (Or `line.clusters_x`, `line.cluster_y`) I don't see any safety checks for the data format prior to this function call, and we'd need some failsafes to ensure the basic transcription functions still work if, e.g. Caesar inadvertantly changes the data format.

src/lib/construct-line.js
```diff
@@ -0,0 +1,46 @@
...
+        />
+      );
+    }
+  });
+  return lines;
+}
+
+function constructPoints(line) {
+  const points = [];
+  line.clusters_x.map((point, i) => {
```
shaunanoordin: Same comments as above - we're not performing any existential checks on `line`, `line.cluster.x` and `line.cluster_y`, so there's a possibility for a crash due to external data format wonkiness.

src/lib/construct-line.js
```diff
@@ -0,0 +1,46 @@
...
+        cx={x} cy={y} r={5} fill="#c33"
+      />
+    );
+  });
+  return points;
+};
+
+function constructText(line) {
+  const sentences = [];
+  line.clusters_text.map((value) => {
```
shaunanoordin: Same comments as above - we're not performing any existential checks on `line` or `line_text`

src/components/AnnotationsPane.jsx
```diff
@@ -52,7 +53,7 @@ class AnnotationsPane extends React.Component {
         svgPoints.push(
           <circle
             key={svgPointPrefix+i}
-            cx={point.x} cy={point.y} r={10} fill="#c33"
+            cx={point.x} cy={point.y} r={10} fill="#00CED1"
```
shaunanoordin: Now that we're more closely following the colour rules of Issue #64 instead of using placeholder colours, we should (soon) take a look at storing these variables into something like, say, `src/app-colours.js`, or use CSS to define these colours. (Disclaimer: my CSS-fu on SVG elements isn't that fantastic.)

wgranger: That makes sense. I don't think you're allowed to use inline constants with HTML, so moving it into CSS would be best. We also need to figure out a way to notify components that an annotation is selected, as that annotation line will change to a green color.

src/lib/construct-line.js
```diff
@@ -0,0 +1,46 @@
...
+    value.map((word, i) => {
+      if (!sentences[i]) { sentences[i] = []; }
+      if (word.length) { sentences[i].push(word); }
```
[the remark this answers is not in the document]
wgranger: I'm not sure how a user would be able to pass in `null` here, and I'm trying to think of how this could potentially appear in `clusters_text`. As far as a string being `'word.length'`, I think that would be fine if that's what a user enters as a classification (if I understand this question correctly).

src/lib/construct-line.js
wgranger: Ahh, good catch.

wgranger:
Ahh, you know what. When I initially started working on this PR I had intended on clicking on a previous annotation to submit a text annotation (with the "done" button in the pane). However, I later decided to submit this PR with the line additions from Caesar's extractor and then deal with annotation submission afterwards. 

Either way, I don't like merging this PR with an obvious bug such as the `TypeError` you are experiencing above. I'll double check that error when covering some of these PR comments.

wgranger:
I'm wondering if any more changes need to be made to the extractor since that example image you posted above contains text from three lines. We could tell users to only mark in the middle or bottom of a text, but inevitably, someone lines could be combined if that is already happening from the extractor.

wgranger:
Alright. I made a good bit of changes amounting to adding around 200 lines and deleting about the same. Mainly, there was repetitive behavior between `previousAnnotations` and `annotations`. Also, some of the naming conventions were strange.

New functionality:
- Saving collaborative annotations
- Changing selected lines green
- Allowing `previousAnnotations` to be seen with frame change

shaunanoordin:
The latest changes have been noted; I'm performing a review of the updates and later a re-review of the whole PR.

src/components/AnnotationsPane.jsx
```diff
@@ -103,16 +104,50 @@ class AnnotationsPane extends React.Component {
...
+
+  renderAnnotations(annotations, previousAnnotations = false) {
+    if (!annotations) return null;
     if (!this.props.showPreviousMarks) return null;
 
     const annotationPrefix = 'ANNOTATION_';
 
-    return this.props.annotations.map((annotation, indexOfAnnotation) => {
+    return annotations.map((annotation, index) => {
+      if (annotation.hasCollaborated === true) { return null; }
```
wgranger: I decided to add the `hasCollaborated` tag on previous annotations to toggle to true once a user has selected and saved a text annotation from the dropdown. This is because, once a user interacts with a previous annotation from the dropdown and saves text, that becomes a **new** user annotation. 

This line would return `null` if a previous annotation has already been "collaborated with" so the renderer doesn't attempt to display a user annotation and previous annotation on the exact same line.

src/components/AnnotationsPane.jsx
```diff
@@ -13,14 +13,15 @@ import React from 'react';
...
-import { constructLines, constructPoints, constructText } from '../lib/construct-line';
 
 class AnnotationsPane extends React.Component {
   constructor(props) {
     super(props);
     this.renderAnnotationInProgress = this.renderAnnotationInProgress.bind(this);
     this.renderAnnotations = this.renderAnnotations.bind(this);
     this.renderPreviousAnnotations = this.renderPreviousAnnotations.bind(this);
+    this.renderUserAnnotations = this.renderUserAnnotations.bind(this);
+    this.determineGreenLine = this.determineGreenLine.bind(this);
```
shaunanoordin: [minor] At a glance, what a 'GreenLine' is isn't too clear, but to be fair, our logic's already pretty complex no matter what.

src/components/AnnotationsPane.jsx
```diff
@@ -103,16 +104,50 @@ class AnnotationsPane extends React.Component {
   /*  Renders all the annotations that the user has completed.
       WARNING: Not to be confused with annotations from other users!
    */
-  renderAnnotations() {
+  renderUserAnnotations() {
```
shaunanoordin: 👍

src/components/AnnotationsPane.jsx
wgranger: Yeah, I spent a while trying to decide the best way to do this. With the streamlined `annotations` and `previousAnnotations`, it was a bit tricky to discern which annotation was selected and should display green. I'd be open to changing the variable name if you think that'd help readability, unless you have other suggestions.

src/components/FilmstripViewer.jsx
```diff
@@ -18,6 +19,7 @@ class FilmstripViewer extends React.Component {
 
   changeFrame(i) {
     this.props.dispatch(changeFrame(i));
+    this.props.dispatch(changeFrameData(i));
```
shaunanoordin: I assume that, without this call, when you change from Frame 1 to Frame 2, Frame 1's Previous User Annotations and/or Aggregated Annotations From Other Users would still show?

wgranger: Actually, I noticed with a change in frame, the previous annotations would just disappear. There was probably some line that returned `null` or something if the data frame didn't match the current subject viewer frame.

shaunanoordin:
## PR Re-Review
- This PR updates the Annotations Pane to properly display the content of Aggregated Annotations (i.e. "Other People's Previous Annotations", taken from Caesar)
screen shot 2017-10-12 at 18 49 37
_In this screenshot, the top two (teal) Annotation lines were created by the user. The third (green) Annotation line (the Selected Annotation) is actually an Aggregated Annotation; by selecting it, we see a text input modal with a dropdown showing 3 text transcriptions from other users._
- The newest commits to this PR addresses a few items:
  - Current/Selected (a new data item I'm adding/selecting), Previous (data items I've added), and Aggregated (data items from other people) Annotations are now clearly indicated by the colours green, teal, and red, respectively.
  - A user may once again click on a Previous Annotation to edit what she/he had written. (This feature was either earlier absent or inadvertently reverted.)
  - **A user can click on an Aggregated Annotation to agree/disagree/edit with other users** - clicking on a red Annotation will open the Text Input modal _with a dropdown listing "other people's transcriptions" to choose from_
    - When the user submit her/his agreement/disagreement/edit, the **Aggregated Annotation (red) visually transforms into a Previous Annotation (teal).**
    - Behind the scenes: _a new Annotation is added to the user's Annotations,_ while the corresponding Aggregated Annotation is simply _hidden visually._
    - This also means that once an agreement/disagreement/edit is made, the "other people's transcriptions" dropdown disappears for that line of Annotation.
    - **TODO:** check what happens when we delete an Annotation that was based on an Aggregated Annotation. Expected: the previous Aggregated Annotation (and its "other people's transcriptions" dropdown) should reappear. 🎏 
    - **NOTE:** The Classifications submitted to Panoptes **do not** yet track whether a user agreed/disagreed with another user for each Annotation. This is a near-future TODO. 🎏

### Status
The code LGTM and functional tests run OK. The latest commits build on top of an already solid PR and address most - if not all - the notes that were brought up in the previous review.

This PR is merge-ready, but I'll run one final pass through the code to ensure I didn't miss anything major. Expect an approval + merge in 15 minutes.

shaunanoordin:
For other devs reading this: please note that when I say "Aggregated Annotations" in my review, the code is sitting there in `ducks/previousAnnotations`, which is different from "My Current/Previous Annotations" (which in turn sits in `ducks/annotations`). There's some terminology to be reconciled, hope you aren't confused!

PR #83, opened 2017-10-12
Consolidate app config

shaunanoordin:
## PR Overview
- This PR improves the app config:
  - the config now exports the `env` environment for other components to reference.
  - Note that I didn't pull `src/lib/split-config.js` into `src/config.js` as I'm unsure if there's any further split-specific code that needs to go in there. If not, it's a good candidate to consolidate into the main config.js.
- [minor] Removed the activeClassName from a few <a> elements. <a> elements cannot have such an attribute.

### Status
Should be ready for review, but I'm doing one more pass as I have a nagging feeling I missed out on a few env-reliant code that could be consolidated.


shaunanoordin:
## PR Update
- The project-links file has now also been consolidated into the app config.

### Status
- Ready for review.

wgranger:
This PR looks good and appears to clean up the code a good bit. I'm seeing the error below, but it also appears with how the app was set up beforehand. I'm going to merge this PR as is and open an issue with the error statement below to revisit later in an attempt to improve app performance. 

screen shot 2017-10-12 at 11 53 22 am

PR #85, opened 2017-10-12
Add header

wgranger:
This PR uses the shared `Header` component from Zooniverse-React-Components. While an earlier commit of this branch contained some functionality for a modal popup with Google sign-in, some discussion is still needed to flesh out the login procedures for the app. The branch should be stable and working.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "redux-thunk": "^2.1.0",
     "seven-ten": "^3.1.0",
     "zoo-grommet": "^0.2.1",
-    "zooniverse-react-components": "github:zooniverse/Zooniverse-react-components#v0.4.3"
+    "zooniverse-react-components": "github:zooniverse/Zooniverse-react-components#v0.5.0"
   },
   "devDependencies": {
     "autoprefixer": "^6.5.3",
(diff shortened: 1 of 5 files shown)


shaunanoordin:
## PR Review
- This PR replaces the custom `<header>` we use with the proper (and shiny, new) shared Zooniverse Header from the ZRC repo.
- Functionally, this PR passes OK - the sign in functionality works, and the links are correct.
- Visually, the header looks like a proper Zooniverse header, but there are some slight quirks to note. (See below)
- Tested OK on OSX+Chrome61/FF55/Safari, Win7+IE11.

### Visual Quirks
This is how our (custom) Zooniverse Header looks like:
screen shot 2017-10-13 at 15 39 52
_Chrome above, Firefox below_

This is how the new ZRC Zooniverse header looks like:
screen shot 2017-10-13 at 15 35 35
_Note the visual difference between the font weights of each browser's Zoo Header._

While I'm happy to accept a change in style/visual design moving from the custom Zoo Header to the proper ZRC Zoo Header, it's curious to note that the ZRC Zoo Header doesn't have a consistent style across browsers.

Current testing has shown that the **fonts** _look_ bolder/fatter on all tested non-Chrome browsers, but look (correctly) slim only on Chrome. Strangely, the calculated CSS data seems correct on all browsers - so it may be less a CSS issue and more of an oddity of rendering the specific font?

I'm not sure if this is an issue worth investigating, but even if it is, it's not for this PR to fix.

### Status
Approve, merge, go.

PR #88, opened 2017-10-13
Remove Unused Components

wgranger:
There are a couple components sitting around from the initial setup of the app and the starter repo. This PR removes those items. Also, I initially started renaming the `About.spec` to test the AboutTheCollection component, but webpack doesn't like the extensions on some image imports (ex. `import AboutCharles from '../images/about-charles.png';`), so some configuration is needed before testing components with image imports.


shaunanoordin:
screen shot 2017-10-13 at 19 01 28

Note the differences between localhost and production:
- Production's project description is blank. (Everything after "Welcome to Anti-Slavery Manuscripts")
- The font of the main h1 element is very, very different. Everything else seems OK, strangely.

shaunanoordin:
GODDANGIT. Sorry, ignore that previous post. That was supposed to go in a separate issue; I have way too many tabs open.

PR #90, opened 2017-10-13
Seven Ten Views

### LINKED ISSUE #70: Add 7-10
This is necessary for split testing. A 7-10 client needs to be included.

Should split information be saved in the metadata? When a classification is made, information should be sent to 7-10 via the client. This will be a visibility split under the 7-10 client.

**Question**: Should the 7-10 server be hit each time a classification is updated, or only when a user completes a classification (all lines, etc.). I'm thinking the latter, but this is worthy of a conversation.

wgranger:
This should be set up properly, but I am going to double check we are submitting seven-ten classifications correctly. I have confirmed (using two test users) that previous annotations will only appear if a user is in the collaborative variant.

Edit: Switching this to WIP while I confirm seven-ten is queried correctly.

Closes #70

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -20,7 +20,7 @@ class App extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.user != nextProps.user) {
+    if (nextProps.user && nextProps.user != this.props.user) {
       this.props.dispatch(fetchSplit(nextProps.user));
     }
   }
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -20,7 +20,7 @@ class App extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.user && nextProps.user != this.props.user) {
+    if (nextProps.user && nextProps.user !== this.props.user) {
       this.props.dispatch(fetchSplit(nextProps.user));
     }
   }
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -198,6 +198,8 @@ class AnnotationsPane extends React.Component {
         </g>
       );
 
+      if (previousAnnotations && this.props.adminOverride) { return renderedMarks; }
+
       if (this.props.splits && previousAnnotations) {
         return (
           <VisibilitySplit key={annotationPrefix + index} splits={this.props.splits} splitKey={'classifier.collaborative'} elementKey={'div'}>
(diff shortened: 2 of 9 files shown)


wgranger:
This should be all set now. A couple changes needed to made to seven-ten, particularly PR https://github.com/zooniverse/Seven-Ten/pull/20#pullrequestreview-69351399

src/components/App.jsx
```diff
@@ -20,7 +20,7 @@ class App extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.user != nextProps.user) {
+    if (nextProps.user && nextProps.user != this.props.user) {
```
shaunanoordin: Is the `!=` intentional, or is this supposed to be `!==` ?

wgranger: Oh, no it probably needs to be `!==`. Good catch.

wgranger:
Yes, that does make sense about testing, since the frontend will now be restricted to whatever group you're in. I'm thinking to add a toggle button that's available if the user is an admin? I'd probably make it small and inconspicuous, off to the side somewhere.

wgranger:
So, an admin toggle has been added, but I'd like some feedback on it. Perhaps it should only show up in development? It's pretty basic, and I'm open to suggestions.

shaunanoordin:
I like what you did there - putting the the toggle as an admin-only viewable option is good with me. I'm also keen on seeing this on _both_ Development and Production, because we might need that option to debug in the future.

Only two caveats at the moment:
- this only allows admins to who have been assigned B (solo) to view see the A (collab) view, but not vice versa.
  - While visually and functionally the same at the user level, (i.e. the only practical difference is whether you see prev annotations or not) the override should affect the code at the logical ("selecting A or B switch") level.
- there's still no indicator if we're on the A or B split.

That said, I'm OK to 👍  this PR and merge it, but if I have time in the future (hahaha) I'd like to revisit the A/B split logic into something like this...

```
function showAorB() {
  if (adminOverride === "on") {
    if (adminOverrideChoice === "A") { return "A"; }
    else if (adminOverrideChoice === "B") { return "B"; }
  }

  if (userIsLoggedIn) return sevenTen_getAorBSplit();

  return "B";  //Default
}
```

```
/*  A tiny, discreet icon on the screen would help us debug a user's situation - when they report an issue, we can ask them to submit a screenshot.
 */
class TinySplitIndicatorIcon extends React.Component {
  ...
  if (showAorB === "A") return <Icon width="5px" height="5px" color="red"></Icon>
  else if (showAorB === "B") return <Icon width="5px" height="5px" color="blue"></Icon>
  ...
}
```

wgranger:
The seven-ten split is currently only active on staging, so another test would need to be set up. However, I am seeing calls to seven-ten being made in the network tab with staging.

wgranger:
This makes sense about switching between collaborative and non-collaborative. I'll open an issue for it.

PR #92, opened 2017-10-16
Feature: Delete Selected Annotation

shaunanoordin:
## PR Overview
- This PR adds the ability to delete an active (Selected) Annotation.
- The Delete button appears on the Text Input dialog that appears when a Selected Annotation is, well, selected. (Either upon clicking an existing Annotation that the user created, or after finishing the process of creating a new one. The Delete button does NOT appear on Aggregated (other people's) Annotations.
screen shot 2017-10-16 at 19 10 01

### Issues to fix
- [x] When you create an Agreement Annotation from an Aggregated Annotation, and then delete said Agreement Annotation, the previous Aggregated Annotation doesn't re-appear. WHOOPSIEDOODLE.

### Thoughts/Possible improvements
- [x] Should clicking on "Delete" prompt a confirmation from the user... or what that be too damn annoying? I think the latter.
- [x] When creating a new Annotation, the Cancel button does _not_ delete the newly created Annotation. But should it? Think about it - user makes three dots, clicks the last one, then thinks "Oh, wait, that was wrong." Should "Cancel" when a line is newly created and has text === '' imply that the user meant to undo creating this line? Or am I worrying too much, since the Delete button does exactly what it needs to?
- [ ] That Delete button could use a visual upgrade. It's too innocuous and can be confused for a Done button. Solution: add icons? Colours?

### Status
WIP. When those checkboxes above are all ticked out, I'll submit this for a full review.

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -3,7 +3,7 @@ import Rnd from 'react-rnd';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { collaborateWithAnnotation, updateText, deleteSelectedAnnotation } from '../ducks/annotations';
-import { updatePreviousAnnotation } from '../ducks/previousAnnotations';
+import { updatePreviousAnnotation, reenablePreviousAnnotation } from '../ducks/previousAnnotations';
 
 const PANE_WIDTH = 800;
 const PANE_HEIGHT = 260;
@@ -167,6 +167,7 @@ class SelectedAnnotation extends React.Component {
   }
   
   deleteAnnotation() {
+    this.props.dispatch(reenablePreviousAnnotation());
     this.props.dispatch(deleteSelectedAnnotation());
     this.props.onClose();  //Note that deleteSelectedAnnotation() also runs unselectAnnotation(), but this needs to be called anyway to inform the parent component.
   }
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -131,14 +131,11 @@ class SelectedAnnotation extends React.Component {
           </div>
 
           <div className="selected-annotation__buttons">
-            {(this.props.annotation.previousAnnotation) ? null :
-              <span>
-                <button onClick={this.deleteAnnotation}>DELETE DELETE DELETE</button>
-                {' | '}
-              </span>
-            }
             <button onClick={this.saveText}>Done</button>
             <button onClick={this.props.onClose}>Cancel</button>
+            {(this.props.annotation.previousAnnotation) ? null :
+              <button onClick={this.deleteAnnotation}>Delete</button>
+            }
           </div>
         </div>
       </Rnd>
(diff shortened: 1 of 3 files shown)


shaunanoordin:
## PR Update
- The Delete button, when deleting an 'Agreement Annotation', now properly reenables the Previous Annotation that the Agreement Annotation was based on. 

### Open questions
- Hmm, I'm not sure if the 'brute-force-match' method is the best way of determining which Previous Annotation (which needs to be reenabled) matches the Selected Agreement Annotation (that's being deleted)
  - Actually, the best way to do this is to make sure all Agreement Annotations has a **pointer** to the Previous Annotation they're agreeing with...
  - BUT that requires a fair amount of reworking of the existing code - the whole Annotations structure (and possibly the resulting Classification being submitted).
  - I'm wondering if it'll be better to do this now, or punting it for the future.
- Any thoughts on the visual design? If minor, I'll take of it now, otherwise, I'll move it to a latter PR.

### Status
Ready for review and/or comments, tagging @wgranger . I'm happy to rework this PR if necessary.

wgranger:
Yeah, I see what you mean with the brute force method. It's not ideal. I thought about adding some sort of attached index when writing the code to make annotations, but I abandoned it because I couldn't justify the purpose for the logic involved. I suppose this is the purpose. My initial thought was to attach the new annotation's index to that previous annotation. However, this will take some thought, so maybe it is worth saving for a future PR.

This annotation box design should be looked over first by @beckyrother since I know that annotation box has been redesigned. I know the "cancel" button has been removed, so for the purposes of this PR, it's really only the addition of the "delete" button.

Bug: There's a notable bug I noticed while working through this PR. If you agree with a previous annotation (making your own, new annotation) and navigate to another frame, when you return the color will not be cyan. I'll have to investigate this on another branch, but it's not related to what is in this branch.

beckyrother:
So sorry!! Removing the "cancel" button was a mistake on my part. Please keep it in there. 

@shaunanoordin is right, though, the buttons do need visual differentiation. 

What does the Delete button do?

Please fill in the Done button and keep Cancel (and Delete?) the outline style.

shaunanoordin:
So @beckyrother accidentally did a delete of cancel, which is incorrect, so we need to cancel the delete of cancel and make sure we don't cancel out delete? OK, done.

shaunanoordin:
The new Delete button removes a line Annotation - i.e. it deletes the circles, lines, and text that the user just created, or just selected. Also, 👍 on filling it with the outline style.

beckyrother:
Hahaha thanks for clearing that up Shaun.

Is that Delete functionality something that's standard in transcription projects? For collaborative at least, it seems like it might be dangerous. Will it have an "are you sure?" modal or confirmation?

shaunanoordin:
Yup, the Delete functionality is pretty standard - you can find it on [Shakespeare's World](https://www.shakespearesworld.org) for example, where clicking on an Annotation you just created shows an Edit/Delete option.

The logic is that the Delete button is there to correct _your own_ mistakes - fortunately, you can't go around deleting what _other people_ transcribed! 👌 As such, I don't see the Delete function as too dangerous (but then again, _I'm a man who likes to live life on the edge_ 🕶 ) but no worries, we can add in a Confirm Y/N dialog.

I would only resist adding a Y/N Confirm dialog if it was attached to an action that the user performs repeatedly, and frankly the user shouldn't be making so many mistakes that they'd need the Delete option at rapid fire, unless s/he's part of some Three Stooges routine.

beckyrother:
Ok, great! As long as it's not possible to delete _other people's_ transcriptions, I'm on board!

PR #93, opened 2017-10-16
Get Recent Subjects on Home Page

wgranger:
This PR adds functionality to retrieve recent subjects from a project's talk board.

### DIFF
--- a/src/components/SocialSection.jsx
+++ b/src/components/SocialSection.jsx
@@ -1,17 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Divider from '../images/img_divider.png';
+import { getSubjectLocation, getThumbnailSource } from '../lib/get-subject-location';
 
-const SocialSection = ({ classifications, completedSubjects, percentComplete, project, subjects, volunteers }) =>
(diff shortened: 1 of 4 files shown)


src/components/SocialSection.jsx
```diff
@@ -1,17 +1,31 @@
...
-    <h2 className="main-title">Subjects<i>of</i>  Note</h2>
-    <img role="presentation" className="divider" src={Divider} />
-    <div className="home-page__subject-row flex-row">
-      <div></div>
-      <div></div>
-      <div></div>
-      <div></div>
-    </div>
+
+    {subjectsOfNote.length && (
```
shaunanoordin: Will `apiClient.type('subjects').get(newestSubjects)` ever return a `null` ? If so, this should be `{subjectsOfNote && subjectsOfNote.length && {`

shaunanoordin:
Due to a weird, weird issue with GitHub, @wgranger actually submitted a [fix](https://github.com/zooniverse/liberating-the-liberator/commit/7b4a518f85b3c4d47c9e6e578a247ed4bc1740ed) for the `subjectsOfNote` existential check, but that commit isn't appearing in this PR. 

In this case, I'm going to approve+merge this PR, then that quick commit can be re-applied separately.

PR #94, opened 2017-10-16
Color Grey Consensus Line

### LINKED ISSUE #64: Determine Dot Colors When Selected, etc.

wgranger:
This PR was done pretty quickly based on some conversation about reaching a line consensus. In this branch, a line will have reached consensus if the consensus score is three or above. When that happens, the onClick for the line disappears and the line appears grey. I believe this is the last item needed for issue #64.

Closes #64

### DIFF
--- a/src/ducks/previousAnnotations.js
+++ b/src/ducks/previousAnnotations.js
@@ -99,6 +99,7 @@ const constructAnnotations = (reductions, frame) => {
         const textOptions = constructText(annotation, i);
         const data = {
           points, frame, textOptions,
+          consensusReached: annotation.consensus_score >= 3,
           previousAnnotation: true,
           hasCollaborated: false,
         };
--- a/src/ducks/previousAnnotations.js
+++ b/src/ducks/previousAnnotations.js
@@ -1,5 +1,6 @@
 import { request } from 'graphql-request';
 import { constructCoordinates, constructText } from '../lib/construct-previous-annotations';
+import { CONSENSUS_SCORE } from '../config.js';
 
 const initialState = {
  data: null,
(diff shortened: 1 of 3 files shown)


src/ducks/previousAnnotations.js
```diff
@@ -99,6 +99,7 @@ const constructAnnotations = (reductions, frame) => {
         const textOptions = constructText(annotation, i);
         const data = {
           points, frame, textOptions,
+          consensusReached: annotation.consensus_score >= 3,
           previousAnnotation: true,
```
shaunanoordin: Side note: in a future refactor, I'll want to revisit boolean variable names and make them a bit clearer in regards to their true/false nature, e.g. "previousAnnotation" -> "isAPreviousAnnotation". This is a personal note and doesn't affect this PR since `consensusReached` makes enough sense as is. (Unless, dare I suggest... `hasSufficientConsensus` ? `isFullyConsensusified` ?)

src/ducks/previousAnnotations.js
```diff
@@ -99,6 +99,7 @@ const constructAnnotations = (reductions, frame) => {
         const textOptions = constructText(annotation, i);
         const data = {
           points, frame, textOptions,
+          consensusReached: annotation.consensus_score >= 3,
```
shaunanoordin: Recommendation: move the value `3` to the config, so we can adjust it more easily in the future.

shaunanoordin:
The latest updates have addressed my comments regarding putting the consensus score into the config. Nice! 👍

PR #101, opened 2017-10-19
Redesign Annotations Pane

wgranger:
A few small design changes were made to the annotations pane as well as the addition of the "underline" text modifier (whose functionality will need to be added soon).

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -99,22 +99,28 @@ class SelectedAnnotation extends React.Component {
         <div className={ENABLE_DRAG} ref={(c) => {this.dialog = c}}>
           <div>
             <h2>Transcribe</h2>
-            <button className="fa fa-close close-button" onClick={this.props.onClose}></button>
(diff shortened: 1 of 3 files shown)


wgranger:
1) The number was removed in the most recent design update.
2) There might've been another branch merged recently that affected the styles on this, so I'll have to double check.

shaunanoordin:
Thanks for clarifying the design update, @wgranger 👍 

OK, I've tested this across OSX+Chrome62/FF56/Safari11 and Win7+IE11 - it looks good across all browsers, save for one exception: OSX+Chrome62, when viewing a Previous Annotation.

screen shot 2017-10-23 at 18 17 31
_OSX+Chrome62, when clicking on a Previous Annotation._

screen shot 2017-10-23 at 18 17 51
_Zoom. Enhance._

The dropdown icon looks a bit off - like it was offset by just a little bit. And only for Chrome. Eh?

That said, I'm thinking that this PR is good to merge. The visual error is minor enough that it can be fixed in a later PR, and for the moment, we have other things to focus on.

PR #102, opened 2017-10-19
Field guide

wgranger:
I threw this together pretty quickly, so I'm open to suggestions. Note: I removed the resize functionality on the dialog popup because resizing the field guide would cause the next page of hidden items to become visible to the user (bad UX).


src/ducks/field-guide.js
```diff
@@ -0,0 +1,51 @@
+import apiClient from 'panoptes-client/lib/api-client.js';
+import { config } from '../config';
+
+const FETCH_GUIDE = 'FETCH_GUIDE';
+
+const initialState = {
+  icons: null,
+  guide: null
```
shaunanoordin: One thing to note is that the Field Guide doesn't follow the standard `status: idle/fetching/success/error` pattern of our other ducks.

On one hand, this will make understanding the state of the Field Guide - e.g. for telling the user the status of the guide - more difficult.

On the other hand, the Field Guide is a low enough on the priority list that I won't make too much of an issue for it. Again, this is something we can improve in the future - we just need the MVP for the beta test at the moment.

src/components/FieldGuide.jsx
```diff
@@ -0,0 +1,129 @@
...
+  deactivateCard() {
+    this.setState({ activeCard: null });
+  }
+
+  renderItem(items, index) {
+    return (
+      <div className="field-guide" key={`FIELD_GUIDE_PAGE_${index}`}>
+        {items.map((item, i) => {
+          const resource = item.icon;
+          const src = this.props.icons[resource].src;
```
shaunanoordin: IE11 spits out an error here - "Unable to get property 'src' of undefined or null reference".

Nyeeerp? Sounds like IE11 has an issue with the... resource... key? Whatever, easiest fix here is to apply a quick existential check before assigning the value of `src`

shaunanoordin: The error happens in Edge too, so an existential check seems required for the Field Guide to work.

src/components/FieldGuide.jsx
```diff
@@ -0,0 +1,129 @@
...
+          )
+        })}
+      </div>
+    )
+  }
+
+  renderActiveCard() {
+    if (!this.state.activeCard) { return null; }
+    const card = this.state.activeCard;
+    const src = this.props.icons[card.icon].src;
```
shaunanoordin: Nyarp, this may require an existence check as well.

src/ducks/field-guide.js
wgranger: Nah, I need to start doing this as some API calls are getting murky in their dependencies of other resources. I'll try to make the change pretty quickly so I can get comfortable with the approach and apply it to other items.

shaunanoordin:
I just noted that this branch has been updated with a rebase, and I think some IE11-fixing commits? Running a second review now.

PR #103, opened 2017-10-19
On successful Classification submission, reset the views/data then fetch next Subject

shaunanoordin:
## PR Overview
- With this PR, when a user _successfully_ submit a Classification, the app's data state is wiped clean, in preparation for the next Subject to be fetched.
- Alas, this is a bit difficult to test since the Subject queue I see on `localhost+zootester1` at the moment only has one Subject that repeats ad infinitum.
- I do see that the Annotations I made are correctly reset when the next (same) Subject is (re-)fetched, so that's good.

### For testing
Expected actions and results:
- User makes an Annotation on the Subject.
- User clicks DONE.
- Expected: there should be a POST to Panoptes's /classifications
- Expected: the Annotation that was made should disappear.
- Expected: there should be several GETs, to Panoptes /subjects and to the Caesar aggregation.
- Expected: a new Subject should appear (with any Previous Annotations for that Subject, if applicable.)

### Thoughts
- Submit Classifications in the Classifications duck was refactored - the Reducer now only performs data updates; all logic (esp. async logic) has been moved out to the Action creator.
- This PR doesn't really have a good handle of what to do when the Classification submission fails, however. Just an ugly alert... this goes on the ToDo.
- There's an open question of _what should I reset before the next Subject?_ e.g. should I run resetSubject() then fetchSubject(), or should fetchSubject() _implicitly_ reset the Subject data? Hmm, hmm.
- The placement of the reset-then-fetch-next-Subject logic is... uh... I feel there's something odd about it being in the Classifications reducer. I mean, it works, but my internal editor is shaking his head telling me that the flow of logic should be easier to see, at a glance, by other developers. His disdainful expression betrays no tidbits of advice, however.

### Bonus
- Temporary Workflow/Subject IDs are now placed in the config.
  - Wait, didn't I do this before? I have a feeling of deja-vu.

### Status
Ready for review, @wgranger

PR #105, opened 2017-10-20
Functioning Subject Set Selection

wgranger:
This PR allows a user to select a subject set to annotate from the landing page. Hardcoded workflows and subject IDs were removed, and subject set info was added to the config.

**To consider before merge:** once this is merged, it will be more difficult to find a subject with previous annotations when debugging. Perhaps some sort of admin toggle should be added? Open to suggestions.


src/components/Home.jsx
```diff
@@ -80,11 +86,15 @@ class Home extends React.Component {
           <img role="presentation" className="divider" src={Divider} />
           <div className="home-page__body-text">
             <b className="body-copy-first-word">Welcome.</b>{' '}
-            {this.props.project.description}
+            {this.props.project && this.props.project.description}
```
wgranger: Existential checks were also added in a couple places in case the project has not yet been retrieved.

shaunanoordin: 👍

src/config.js
```diff
@@ -27,6 +27,7 @@ const baseConfig = {
       host: 'https://master.pfe-preview.zooniverse.org/',
       projectId: '1764',
       projectSlug: 'wgranger-test/anti-slavery-testing',
+      workflowId: '3017'
```
shaunanoordin: PR #103 also adds in [Workflow ID](https://github.com/zooniverse/liberating-the-liberator/pull/103/files#diff-9d9a6cd82f41984872a66a3ab0d440c4R31) so a note has to be made during a PR rebase (depending if either 105 or 103 is merged first).

shaunanoordin:
> To consider before merge: once this is merged, it will be more difficult to find a subject with previous annotations when debugging. Perhaps some sort of admin toggle should be added? Open to suggestions.

Question: why would it be harder to find Subjects that have had aggregated results? On localhost, isn't the default Workflow ID still 3017, and the Subjects there have aggregated results?

wgranger:
Yes, the default workflow ID is still 3017. However, the test image we were using (with previous annotations) was from a subject set that isn't included in those year groups. In order for subject set selection to work, each subject set has to be activated on the 3017 workflow. With all subject sets activate, you're at the luck of the draw on subject sets when you choose "Transcribe Random," and the other subject sets I created for purposes of testing this branch only have one image which has no classifications attached to it. Thus, no previous annotations will be returned until classifications are made on those images.

Hope that makes sense.

shaunanoordin:
👍  Alright, thanks to @wgranger re-enabling some Panoptes settings, I can confirm that the Subject Set selection works.

> Slack conversation, wgranger
> Subject set selection only works if the subject sets are attached to the workflow. I just disabled them for that workflow when testing something. If I turn them back on right now for 3017, you should see a difference.

This PR is now ready to merge. 👍

PR #112, opened 2017-10-23
Functioning Text Modifiers

### LINKED ISSUE #65: How Will We Account for Insertions/Unclear/Deletions?
This should be fairly easy to implement, but we need to find out a way to do it data-wise. I'm currently not sure the best solution in notifying a researcher of this with a classification.

Image below for reference:
screen shot 2017-09-08 at 10 27 55 am

wgranger:
This PR contains basic functionality for text modifiers. I borrow code based on suggestions by @simensta. Everything appears to be working; however I still have a couple questions.

1) Basic UI - should we find a way to display this information when it comes back to a user from a previous annotation? ex. Would "This is [insertion]a test[/insertion]" rattle a user?

2) Should we write another bit of code to pad text correctly or trust a user to follow directions? ex. "Here is some [insertion] text I am [/insertion] writing" would throw off the reducer's word count.

Closes #65

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -89,10 +89,16 @@ class SelectedAnnotation extends React.Component {
       }
     }
 
-    const regex1 = new RegExp("\[" + textTag + "\]\s*", "ig");
-    const regex2 = new RegExp("\s*\[\/" + textTag + "\]", "ig");
-    value.replace(regex1, `[${textTag}]`).replace(regex2, `[/${textTag}]`).replace(/\s+/g, ' ');
-    this.inputText.value = value;
+    const startReg = `(?=\\w*)\\[${textTag}\\]\\s*`;
+    const endReg = `\\s*\\[\\/${textTag}\\](?=\\w*)`;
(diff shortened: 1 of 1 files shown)


shaunanoordin:
I still need to go through this PR, but to answer your questions:

1. I'm going to trust that our users will be smart enough to understand `the use of [insertion]raw metatags[/insertion] in the textbox`

2. This should be simple to do in theory, but all the edge cases make me go hmm. One implementation:

`"Sample  [tag]  tagged   [/tag]  text".replace(/\[tag\]\s*/ig, '[tag]').replace(/\s*\[\/tag\]/ig, '[/tag]').replace(/\s+/g, ' ');`
=>
`"Sample [tag]tagged[/tag] text"`

wgranger:
I dove deep into regex to make a couple modifications to the suggestion above (btw, thanks for that bit of code to get started). 

1) The regex _also_ pads the beginning of a start tag (ex: `" [insertion]"`) and the end of a closing tag (ex: `"[/insertion] "`).
2) In the case of an "unclear" text modifier, the modifier is padded at the beginning and end (ex: `" 
[unclear] "`).

PR #113, opened 2017-10-23
Fix Submit Classification bug

shaunanoordin:
## PR Overview
Issue: when a user Submits a Classification (i.e. clicks Done), the next time they do so, they encounter an error.

Analysis:
- the createClassification() initialiser isn't properly called on all instances of a new Subject being loaded.
- See ClassifierContainer:
```
  if (nextProps.currentSubject && nextProps.workflow && !this.props.classification) {
  }
}
```

Solution:
- We force a createClassfication() call within the Submit Classification action itself.
- This may cause double work (since the componentWillReceiveProps seems rather random on when it creates a Classification) but it's a relatively safe method.

### WARNING
- While this PR fixes the Submit a Classification bug, it may expose a problem with the next set of Previous Annotations not being fetched.
- Unfortunately, I can't tell if the `fetchAnnotations()` function simply isn't called, or if it IS called but I'm getting a Subject with no Annotations.
- I'm waiting for some feedback on this.

### Status
Ready for review, @wgranger . Please take note of the Warning note above; I'd like to know what you see on your end.

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -106,23 +106,21 @@ const submitClassification = () => {
     const subject_dimensions = (subject && subject.imageMetadata) ? subject.imageMetadata : [];
     const classification = getState().classifications.classification;
     
-    if (classification) {
-      dispatch({ type: SUBMIT_CLASSIFICATION });
-      classification.annotations.push(annotations);
-      classification.update({
-        completed: true,
-        'metadata.finished_at': (new Date()).toISOString(),
-        'metadata.viewport': {
-          width: innerWidth,
-          height: innerHeight,
-        },
-        'metadata.subject_dimensions': subject_dimensions || [],
-      })
-      .save();
-    } else {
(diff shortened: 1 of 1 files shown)


wgranger:
A potential way to test this is to add another subject to that workflow and submit classifications a few times to see if the previous annotations clear themselves out and are refetched when a previously annotated subject appears again later.

wgranger:
LGTM. Based on my comment above, after submitting a few "classifications" (clicking on the "done" button a few consecutive times) I am seeing another subject appear without previous annotations (a cleared out state) and another subject later being fetched with valid previous annotations. I take that as a sign things are working as expected, but let me know if you have any reservations before I merge this PR.

shaunanoordin:
👍  Nice! Thanks @wgranger for testing that, you've assuaged my concerns. Awright, since it's a LGTY, I'll go ahead and merge. 👌

PR #115, opened 2017-10-24
Complete Tutorial Fetching and Display

wgranger:
The tutorial component from ZRC is now behaving as expected and tutorials are automatically displayed based on a user's preferences (if he/she has seen a tutorial previously or not). 

Note: This branch is titled `view-tutorials-redo` because a bad rebased caused me to ditch the original branch. This is the correct version of that branch that reflects the current master.


wgranger:
Ahh, I see. For starters, guests _are_ given a preferences resource. In this PR the assumption is that they would get `null`. The first issue is prob a fix needed in ZRC.

wgranger:
Ok, this should be good now; however, a couple changes will need to be made on ZRC and then updated here, so we'll wait until those are in.

shaunanoordin:
Acknowledged that Issue 1: 'Continue' button needs a fix in the ZRC.

Current tests show that Issue 2 has been solved, nice. 👍  Logged in and non-logged in users can close the Tutorial popup just fine.

However, Issue 3 seems extant, at least for `zootester1` let me dive in to see what's up.

src/containers/ClassifierContainer.jsx
```diff
@@ -62,6 +65,14 @@ class ClassifierContainer extends React.Component {
     if (nextProps.currentSubject && nextProps.workflow && !this.props.classification) {
       this.props.dispatch(createClassification());
     }
+
+    if (nextProps.workflow && nextProps.preferences && nextProps.tutorialStatus === TUTORIAL_STATUS.IDLE) {
+      this.props.dispatch(fetchTutorial(nextProps.workflow));
+    }
+
+    if (nextProps.tutorial !== this.props.tutorial) {
+      Tutorial.startIfNecessary(Tutorial, nextProps.tutorial, nextProps.user, this.props.preferences);
```
shaunanoordin: Two things:
1. Is there a reason this uses `this.props.preferences` instead of `nextProps.preferences`?
2. Does the _if-clause_ here need to check for the existence of nextProps.tutorial/user/preferences? I'm unfamiliar with the mechanisms of ZRC's Tutorial, alas.

shaunanoordin:
I've added a note regarding the `Tutorial.startIfNecessary()` bit. Unfortunately, I can't really debug much of the ZRC's `Tutorial` component, so I can't do much to figure out Issue 3. (So I guess this goes under the _"fix it in ZRC"_ category?)

That said, my main concern is with making sure that A. users can view the tutorial and B. users can close the tutorial. This PR now fully addresses both primary concerns, and everything else is an issue of fine-tuning.

Unless you want to fix Issues 1 and 3 in this branch, @wgranger , I'm happy to approve and merge this PR and wait for an update in the ZRC.

src/containers/ClassifierContainer.jsx
wgranger: Glad you're bringing this up because I did a good bit of modifying with `this.props` vs. `nextProps`. 

1) Yes, this should probably be `nextProps`.
2) I don't think it needs to check the presence of `user` or `preferences` on this line because it is checked a few lines above for the initial tutorial fetch. A tutorial will not be fetched without preferences or a user, so that `if (nextProps.tutorial !== this.props.tutorial)` line _shouldn't_ be reached unless a user or preferences was already available.

wgranger:
Let's wait for the ZRC fix before this is merged.

wgranger:
Oh! And an update on issue 3. There was a problem with the way preferences were saved due to a typo. You can see the fix in ZRC [here](https://github.com/zooniverse/Zooniverse-React-Components/pull/42/files#diff-62cef40c1a6243b00a2be3ed87335c3aR45)

shaunanoordin:
The ZRC has been updated and I'll be checking this PR again with a fresh reinstall.

NOTE: there are currently some conflicts on this branch, and I'll resolve them via the Git editor. @wgranger, please note that I'll be updating this branch, so you'll need to pull the latest if you want to make any further changes.

wgranger:
The tutorial appears now regardless of user preferences because the project resource is loaded first, then the preferences are retrieved before the user has been received due to the optimization change in #120. Thus, guest preferences are always used.

PR #120, opened 2017-10-25
Streamline fetch lifecycle and fix fetching Subject from queue

shaunanoordin:
## PR Overview
- This PR handles two major issues.
- First, the **fetching Subject-Classification lifecycle** has been refactored.
  - Instead of `createClassification()`  triggering inside `ClassifierContainer` as it waits for a new Subject to come in...
  - ... `createClassification()` is now triggered directly after a Subject is successfully fetched within its Duck. i.e., we're removing the 'React component monitoring for changes' bit.
- Second, this PR fixes an issue where you can't get the third Subject from your queue.

> Issue: after submitting the Classification for the second Subject in your queue, you can't load any more Subjects - the system crashes. 
> Analysis: In `fetchSubject()`, there's an if-else fork that either A. fetches a new Subject Queue and the latest Subject on that queue, or B. fetches the next Subject in an existing queue. Path B was set up in such a way that not all the necessary Subject information was passed to the reducer.
> Solution: fixed that.

@wgranger : Wait I just realised this, have we accounted for Subject Queues that are empty? i.e. no next subject?

This PR replace PR #113 which wasn't a good enough a fix, alas.

### Status
@wgranger Ready for review

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -9,6 +9,9 @@ import Header from './Header';
 import ProjectHeader from './ProjectHeader';
 import Dialog from './Dialog';
 
+import { PROJECT_STATUS } from '../ducks/project';
+import { WORKFLOW_STATUS } from '../ducks/workflow';
+
 class App extends React.Component {
   returnSomething(something) { // eslint-disable-line class-methods-use-this
     return something;
@@ -26,6 +29,11 @@ class App extends React.Component {
   }
 
   render() {
+    if (this.props.projectStatus !== PROJECT_STATUS.READY ||
+        this.props.workflowStatus !== WORKFLOW_STATUS.READY) {
+      return <div>Loading...</div>;
+    }  //TODO: Consider what to do for STATUS: ERROR
+    
     const path = this.props.location.pathname;
     const showTitle = path === '/classify';
 
@@ -51,24 +59,35 @@ class App extends React.Component {
 
 App.propTypes = {
   children: PropTypes.node,
-  dialog: PropTypes.node,
   dispatch: PropTypes.func,
   location: PropTypes.shape({
     pathname: PropTypes.string,
   }),
+  //--------
+  user: PropTypes.object,
+  dialog: PropTypes.node,
+  projectStatus: PropTypes.string,
+  workflowStatus: PropTypes.string,
 };
 
 App.defaultProps = {
   children: null,
   dialog: null,
   location: {},
+  //--------
+  user: null,
+  dialog: null,
+  projectStatus: PROJECT_STATUS.IDLE,
+  workflowStatus: WORKFLOW_STATUS.IDLE,
 };
 
 const mapStateToProps = (state) => {
   return {
+    user: state.login.user,
     dialog: state.dialog.data,
-    project: state.project,
-    user: state.login.user
+    //--------
+    projectStatus: state.project.status,
+    workflowStatus: state.workflow.status,
   };
 };
 
--- a/src/containers/ClassifierContainer.jsx
+++ b/src/containers/ClassifierContainer.jsx
@@ -58,12 +58,6 @@ class ClassifierContainer extends React.Component {
 
   //----------------------------------------------------------------
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.currentSubject && nextProps.workflow && !this.props.classification) {
-      this.props.dispatch(createClassification());
-    }
-  }
-
   componentWillUnmount() {
     Split.clear();
   }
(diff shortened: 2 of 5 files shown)


wgranger:
A rebase will be needed here before merge.

wgranger:
LGTM. Ready to merge after rebased.

shaunanoordin:
A rebase is being force-pushed, invalidating previous incarnations of this branch. Please be prepared.

Also, note that I may have found another weird issue where this PR fails on the newly re-enabled "Random Subject Set" option.

shaunanoordin:
@wgranger So, right, at the moment of this writing...
- In the `master` branch, if you submit (click Done) a few times, you'll eventually run into an error (the Subject Queue issue that this PR fixes).
  - This applies whether you're on a specific Subject Set or a random one - there's one Subject Set that has more than 1 item in it that lets you see the error.
- In this `fix-subject-queue-fetch`,
  - If you choose a specific Subject Set from the Home page, you can submit forevermore.
  - If you choose a random Subject Set, you can't even submit once, because _the Workflow ID isn't set._

That's a hint on where the issue might be - the is Workflow ID set in the usual manner if we go with the random route, correct?

shaunanoordin:
Scratch that, this appears to be random - I'm testing on an incognito window.

shaunanoordin:
screen shot 2017-10-26 at 21 45 43
_The error I see; can't seem to find the conditions that trigger it, though._

wgranger:
Are you talking about `workflow version can't be blank`?

shaunanoordin:
Yeah, and BLARGH I now realise what it is.

God freaking dangit... there's a proper chain of events that need to trigger. Create Classification _must occur after Fetch Subject AND Fetch Workflow succeed._ I've only accounted for Fetch Subject, what.

shaunanoordin:
Wait, wait, no, maybe I'm just insane... not sure how this got into the code: `dispatch(createClassification(createClassification()));`

wgranger:
This bring up a thought I had earlier. I believe the way PFE works is that the app waits until all items are collected before proceeding to load anything. I believe that's done in the `ProjectPageControlled` [here](https://github.com/zooniverse/Panoptes-Front-End/blob/06ef8e443a39e54c581e2df302b2681b704b1992/app/pages/project/index.cjsx), which is something we can investigate implementing?

wgranger:
I saw that crazy line above but thought you knew what you were doing and the code seemed to operate fine with it. I should've tested it more in the review. We're both going a bit stir crazy as this project nears completion.

shaunanoordin:
Actually, it's a combination of I'm insane, as well as spot on on the core problem. Take a look at this...
screen shot 2017-10-26 at 21 55 30

Notice that CREATE_CLASSIFICATION triggers _before_ FETCH_WORKFLOW_SUCCESS, leading to an unfortunate incident where `submitClassification()` triggers without having a valid workflow data to construct it.

EDIT: based on what you said, the `ProjectPageControlled` seems to be on the right path. We need to block any action on the page until all resources are loaded. Worth implementing? Yes. Do we have time? Hmm-

shaunanoordin:
## PR Update
- SURPRISE, I modified App.jsx
- The _whole website_ now just shows a "Loading..." message while it waits for both the _Project and Workflow_ Panoptes resources to load.
  - Future TODO: handle error statuses as well.
  - Future TODO: Make that Loading... message look good.

### Status
@wgranger This is now ready for a re-review. The wait-until-all-resources-have-loaded is basically the ultimate way of us correctly chaining the resources that depend on one another. Despite the very, very few lines of code in this PR, it does feel somewhat like a sledgehammer... but it is effective.

shaunanoordin:
Bonus clarification: the reason `master` doesn't have that problem I discovered is because, when the _ClassifierContainer component_ was 'managing' the fetch life cycles, it basically listened for every update and triggered `createClassification()` whenever _all the pieces_ were in place.

```
  if (nextProps.currentSubject && nextProps.workflow && !this.props.classification) {  //Note that it checks for workflow here
  }
}
```

The method of chain-of-fetch-lifecycles method I'm proposing in this PR separates those fetch lifecycles from the components (which I think is a good idea) but the price to pay is that it requires a more acute awareness of the order required of things. (Which has been flarrrgh.)

I should have noted the nextProps.workflow check though, that one was on me to be more aware.

src/components/App.jsx
```diff
@@ -26,6 +29,11 @@ class App extends React.Component {
   }
 
   render() {
+    if (this.props.projectStatus !== PROJECT_STATUS.READY ||
+        this.props.workflowStatus !== WORKFLOW_STATUS.READY) {
+      return <div>Loading...</div>;
```
wgranger: This makes sense from a performance standpoint, but we should prioritize those simple changes (error handling and styling the loading... in time for beta). Thanks for pointing this out with the Future TODO.

wgranger:
I'm going to merge this under the assumption we will address error handling and the loading message in time for beta. It shouldn't be a huge problem to solve, but it is worth discussion.

PR #122, opened 2017-10-26
Design Adjustments

### LINKED ISSUE #63: Build Tutorial UI
This is ready to go once some mock info is put into the test project. The UI should look similar to the attached wireframe.

https://projects.invisionapp.com/share/RZD7KCGSF#/228369302_Field_Guide

wgranger:
This contains some design adjustments based on suggestions. Also, the project background was given a buffer to prevent it from being too short, and the social icons were updated on the landing page.

Closes #63


src/components/SocialSection.jsx
```diff
@@ -6,7 +6,7 @@ import { getSubjectLocation, getThumbnailSource } from '../lib/get-subject-locat
 const SocialSection = ({ project, subjectsOfNote }) =>
   <div className="home-page__social-section">
 
-    {subjectsOfNote && subjectsOfNote.length && (
+    {subjectsOfNote.length ? (
```
shaunanoordin: Out of curiosity, why was the first part of `subjectsOfNote && subjectsOfNote.length` removed?

I've traced the code and it looks functional - subjectsOfNote is unlikely to be `undefined` unless there was a catastrophic reply from the API, so the code is still practically safe, but the lack of that existence check makes it just a little more brittle.

wgranger: This was because the numeral "0" appeared when there was no subjectOfNote being discussed (empty array) as is currently present on the main site. Adding a `false` at the end of this ternary seemed to take care of the problem.

shaunanoordin: Ah, I understand why it's a ternary, it's that weird zero being rendered. Sorry, I should rephrase: why is it `(subjectsOfNote.length) ? something : false` instead of `(subjectsOfNote && subjectsOfNote.length) ? something : false` ?

Again, it's a curiosity question; it's no big deal as I traced the code and it should be safe.

wgranger: Ahh, good point there. I see what you mean, and having the extra check on `subjectsOfNote` definitely would help here. I think I experimented with a couple different ways of doing this and forgot to add back in the `subjectsOfNote && subjectsOfNote.length` once I found a working solution. I can add back in the extra check with a future PR.

PR #124, opened 2017-10-30
Save Sessions on Mount and Classification Submit

wgranger:
This PR handles work associated with creating a session ID on app load and submitting the session ID along with the metadata for a classification. It's possible this work will be needed when saving classifications in localsStorage, so it's better to break it out into two separate PRs.


shaunanoordin:
A merge conflict was just fixed; @wgranger please be sure to pull the latest branch updates before running on local.

shaunanoordin:
Edit: ah, it turns out that this is standard practice, I just wasn't aware of it.

>  sessions are usually saved in PFE along with classifications.

Thanks, Will!

Issue #125, opened 2017-10-30
Queued Subjects don't make sense


shaunanoordin:
## Issue
Queued Subjects from the Panoptes `/queued` endpoint don't make sense in ASM. (Notably, on staging/localhost, while logged in as `zootester1`, and while logged out. Note that the ASM Subject Sets on localhost have very limited data - having only 1 Subject per Subject Set)

**Issue 1:** Inconsistent queue behaviour between Logged In and Logged Out.

Say I have Workflow X, with Subject Set A (containing one photo: A1), Subject Set B (containing one photo, B1), and Subject Set C (one photo, C1).  I have enabled "select specific subject_set" for workflow X.

When I am logged in, and I ask `/queued?workflow=X&subject_set=A`, the response is `[A1.jpg]` which is expected.

Stangely, when I'm _not_ logged in, and I ask `/queued?workflow=X&subject_set=A`, the response is `[A1.jpg, B1.jpg, C1.jpg]` which is _not_ what I expected.

**Issue 2:** When setting Subject Set ID, the existing Subject queue in the app's memory isn't cleared, so you're always stuck with the previous queue until you finish it.

### Status
Issue 1: Priority, requires investigation.

Issue 2: Quick fix incoming, but it'll be **worth monitoring** once we go in testing and have larger Subject Sets (i.e. more Subjects) to test with.

shaunanoordin:
Closed by #122 and zooniverse/Panoptes#2510

Thanks Marten and Cam!

PR #126, opened 2017-10-30
Fix Subject queue not making sense

shaunanoordin:
## PR Overview
- There are two issues with the Panoptes `/queued` request for Subjects. (See Issue #125)
  - One: logged in vs non-logged in get very different queues.
  - Two: when we set a new Subject Set id, the existing queue isn't cleared, so the new Subject Set isn't actually requested.
- This PR fixes Issue TWO only, by resetting the internal queue whenever a new Subject ID is selected.

## Testing is Madness
TL;DR this is difficult to test, but it looks like it works to me.

Anyway. Testing this will be odd given that ASM has about 5 Subject Queues with only 1 Subject each, and given Issue One above. This is the only way I can tell that it works, on localhost, while not logged in: _(It always looks fine while logged in due to the combination of Issue 1 and the small Subject Sets.)_
1. As a non-logged in user, I always see the same Subject when I choose a Subject Set, e.g. "1800-1839". (Expected: since non-logged in users always receive the same queue in the same order (see Issue One), this means I should always get the same first subject in the ever-resetting queue. Prior to this PR: selecting a new Subject Set (e.g. from "1800-1839") from the home page over and over again has the weird effect of processing through the same queue, resulting in a parade of exact Subjects going on over and over again.).
2. As a non-logged in user, if I click Transcribe in the navi, then home, then Transcribe, then home, etc, I see the "processing through the same queue" mentioned above, which is (mostly) expected in this context.
3. As a logged in user, I notice no changes. (Expected - logged-in users only ever get Subject queues with one Subject since each Subject Set has only one Subject. Therefore, their queue is always fresh.)

### Status
Ready for review, @wgranger , but even if this is approved, it'll be worth monitoring this matter when we go into internal/beta testing and have larger Subject Sets to play with.

### DIFF
--- a/src/ducks/subject.js
+++ b/src/ducks/subject.js
@@ -55,7 +55,8 @@ const subjectReducer = (state = initialState, action) => {
 
     case SET_SUBJECT_SET:
       return Object.assign({}, state, {
-        subjectSet: action.id
+        subjectSet: action.id,
+        queue: [],
       });
 
     case FETCH_SUBJECT_ERROR:


wgranger:
LGTM

PR #129, opened 2017-10-31
Annotation Logic Checks

### LINKED ISSUE #75: Annotation input needs sanity check - lock line angles, check word count vs circle count
## Feature Required
- When a user creates a new Annotation (of a line of text), the UI needs to run two sanity checks to ensure the Annotation's data is valid:
  - [ ] After the first two circles are laid out, the next circles in the line must be within an angle of +/- 5 degrees of the first/previous line segment. (Note: the value of the angle is fairly arbitrary, feel free to argue for +/- 1 degree or +/- 30 degrees.)
  - [ ] When attempting to save an Annotation (at the Text Input dialog), the app can only allow the text to be saved if `text.number_of_words === line.number_of_circles - 1`
- Corollaries:
  - [ ] This means a line of text consisting of only one circle (i.e. 0 line segments) is _invalid._
- The rationale:
  - A line of text must be more-or-less straight in one direction; it can't curve in on itself. We're not accommodating CatFace type of transcription anomalies. We also don't want users to use a _single Annotation_ to cover a _whole page_ consisting of _multiple lines_ by _zig-zagging_ the line segments. That is the path of woe and insanity.
  - Each line segment must indicate one word, and each circle must be placed at the beginning of a word (and at the end of the last word) in a line of text. Therefore, the number of these 'position-of-word-indicating' circles MUST be indicative of the number of words being transcribed, plus the end coordinate of the last word.

For further clarification, see the Slack discussion on #imls on Tue 26 Sep 2017, and/or check Sam's notes for the meeting on that same day.

wgranger:
This PR contains three main logic checks:

1) Disable the "done" button on the `SelectedAnnotation` pane until the correct amount of words have been typed.
2) Disallow a user to create an annotation point unless the points are in a line.
3) Pulsate the most recent annotation point as a signifier that it can be clicked to end an annotation.

*Note: The second item was done with a separate component to keep the entire `AnnotationPane` from updating when `setState` was called with a mouse move. Also, I'm open to ideas on the third item as I know SVG animations are not supported by all browsers, and this was the first solution I arrived upon.

Closes #75

### DIFF
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -64,6 +64,7 @@ class AnnotationsPane extends React.Component {
       if (i === this.props.annotationInProgress.points.length-1) {  //Final node: click to finish annotation.
         svgPoints.push(
           <circle
+            id="pulsating"
             key={svgPointPrefix+i}
             cx={point.x} cy={point.y} r={10} fill="#00CED1"
             className="end"
@@ -106,6 +107,15 @@ class AnnotationsPane extends React.Component {
 
     return (
       <g className="annotation-in-progress">
+        <animate
+          xlinkHref="#pulsating"
+          attributeType="CSS" attributeName="opacity"
+          from="1" to="0.2" dur="1s" begin="0s"
+          begin="pulsating.mouseout"
+          end="pulsating.mouseover"
+          repeatCount="indefinite"
+          fill="freeze"
+         />
         {svgLines}
         {svgPoints}
       </g>
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -14,6 +14,7 @@ import PropTypes from 'prop-types';
 import { Utility } from '../lib/Utility';
 import { connect } from 'react-redux';
 import { VisibilitySplit } from 'seven-ten';
+import PendingAnnotation from './PendingAnnotation';
 
 class AnnotationsPane extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


shaunanoordin:
This PR has been updated; a new review is underway.

shaunanoordin:
## PR Review (Update)
- Issue 2: disappearing Preview Line - is _solved_ 👌 
- Issue 3: When opening an existing Annotation, the word count says 0 out of X words - is still ongoing.

### Status
I'm taking a look at the problem. It seems that the text of an existing Annotation is filled at `componentDidMount(),` so logically, `wordCount()` should be added there. Strangely though, when running `wordCount()` at that part of the React lifecycle, the system just cries. I'm guessing it's stuck in an infinite lifecycle update loop or something.

wgranger:
Ohhh. An existing annotation! I was looking at another bug involving selecting previous annotations. I will address this momentarily.

shaunanoordin:
## PR Update
- @wgranger I've refactored the Word Count system in `SelectedAnnotation.jsx`.
  - This fixes Issue 3, so when you select an existing Annotation, the word count is displayed properly, and the Done button is blocked/unblocked correctly.
  - All other word count functionality has been preserved.
- A note on the implementation: the `wordCount()` logic no longer triggers over multiple text update events. Instead, we count the words - and disable the 'Done' button - on the `render()` action only.

### Status
Since I've made significant changes to this PR, I'm assigning review to @wgranger to evaluate the updates.

wgranger:
LGTM

PR #135, opened 2017-11-01
Add Google Analytics and Geordi Logging

wgranger:
Much of the work on this branch is based on work done with a [current PFE PR](https://github.com/zooniverse/Panoptes-Front-End/pull/4130). 

Some GA tracking items are [still pending](https://docs.google.com/document/d/1BkwASryrJmIqF84yc0Ce6fh9xJHTqG154SNTlyQkt7o/edit), but they are dependent on some other additions made to the app. Additionally, some additions will need to be made on PFE.

General test info can be found [here](https://docs.google.com/document/d/14_dJVVIb9nifF8REQTmDaWjFqD-d0wc6jhlGflt9_Mc/edit).

You should be able to confirm click tracking is sent to Google Analytics and Geordi by keeping an eye on the Network tab while navigating the site.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -36,6 +36,10 @@ class App extends React.Component {
   componentDidMount() {
     this.props.dispatch(fetchProject());
     this.props.dispatch(fetchWorkflow());
+    // this.googleLogger.remember({ projectToken: ['wgranger-test/anti-slavery-testing'] });
+    // if (this.props.user) {
+    //   this.googleLogger.remember({ userID: user.id });
+    // }
   }
 
   componentWillReceiveProps(nextProps) {
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -12,10 +12,25 @@ import LoadingSpinner from './LoadingSpinner';
 
 import { PROJECT_STATUS } from '../ducks/project';
 import { WORKFLOW_STATUS } from '../ducks/workflow';
+import GoogleLogger from '../lib/GoogleLogger';
 
 class App extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.googleLogger = null;
+  }
+
   returnSomething(something) { // eslint-disable-line class-methods-use-this
     return something;
+  }
+
+  getChildContext() {
+    return ({ googleLogger: this.googleLogger });
+  }
+
+  componentWillMount() {
+    this.googleLogger = new GoogleLogger;
   }
 
   componentDidMount() {
@@ -81,6 +96,10 @@ App.defaultProps = {
   projectStatus: PROJECT_STATUS.IDLE,
   workflowStatus: WORKFLOW_STATUS.IDLE,
 };
+
+App.childContextTypes = {
+  googleLogger: PropTypes.object
+}
 
 const mapStateToProps = (state) => {
   return {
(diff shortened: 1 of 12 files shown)


src/components/App.jsx
```diff
@@ -12,20 +12,45 @@ import LoadingSpinner from './LoadingSpinner';
...
+
   returnSomething(something) { // eslint-disable-line class-methods-use-this
     return something;
   }
 
+  getChildContext() {
+    return ({ googleLogger: this.googleLogger });
+  }
+
+  componentWillMount() {
```
shaunanoordin: Is there any condition in which we need to worry about the Component un-mounting? Do we need to unsubscribe from the GoogleLogger, or is this extremely unlikely? (e.g. if App unmounts, it means the darn window has closed.)

src/components/ProjectHeader.jsx
```diff
@@ -3,48 +3,79 @@ import PropTypes from 'prop-types';
...
+class ProjectHeader extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.aboutClick = this.aboutClick.bind(this);
+    this.talkClick = this.talkClick.bind(this);
+  }
+
+  aboutClick() {
+    if (this.context.googleLogger) {
```
shaunanoordin: Wait, `React.context` ? Hissss! Unclean, unclean! Legend tells that React.context was forged by Sauron himself in the heart of Orodruin to tempt the hearts of developers. _Its power comes at great cost._

Theatrics aside, I can see where using context _could_ make sense in our case. We generally avoid React.context because it might be patched out in future versions of React, but if we can accept the ASM project's limited lifespan, we can accept the risks of using context.

src/components/SelectedAnnotation.jsx
```diff
@@ -185,6 +200,14 @@ class SelectedAnnotation extends React.Component {
     );
   }
 
+  cancelAnnotation() {
+    if (this.props.onClose) { this.props.onClose() };
+
+    if (this.context.googleLogger) {
+      this.context.googleLogger.logEvent({ type: 'cancel-transcription' });
```
shaunanoordin: Worth checking: does the cancel-transcription logging need to happen _after_ props.onClose() ? I'm not sure if onClose() kills the component and prevents the log from triggering. Probably not - the code block should run properly, but we could do a quick double check.

src/ducks/splits.js
```diff
@@ -2,19 +2,44 @@ import { Split } from 'seven-ten';
...
+  IDLE: 'split_status_idle',
+  FETCHING: 'split_status_fetching',
+  READY: 'split_status_ready',
+  ERROR: 'split_status_error',
+};
+
 const initialState = {
   adminOverride: false,
-  data: null
+  variant: 'individual',
```
shaunanoordin: The `'individual'` and `'collaborative'` strings can probably be set as CONSTANT_VALUES if we're using them in a lot of places. `export const VARIANT_TYPES = { INDIVIDUAL: 'individual', COLLABORATIVE: 'collaborative' }` ?

src/components/App.jsx
wgranger: Ahh! Yes. I need to address this actually and I'm glad you brought it up. The `forget` function needs to be utilized in case of a user change. I will update this branch accordingly.

src/components/SelectedAnnotation.jsx
wgranger: I confirmed this a few times and it appears the call is made in the network tab before the modal closes.

wgranger:
Before beta launch, we do need to confirm the separation of logging between staging and production. It isn't a huge issue now since we can sort logging by date, but that will need to be considered after this branch is merged.

shaunanoordin:
## PR Review (Update)
- New changes have been added to log when users log out. Also, some code cleanup re: constant variables.
- Existing functionality from the previous PR review has been confirmed.

### Status
LGTM! Let's go.

PR #139, opened 2017-11-02
Fix outstanding issues

### LINKED ISSUE #127: Design: Update fonts from Google Fonts
Just looking at the Google Fonts api call on the home page HTML. I added Playfair Display Italic `400i` and I couldn't find a place where Open Sans is used, so that's been removed. To summarize:

Here's the updated URL:
https://fonts.google.com/selection?selection.family=Cormorant|Playfair+Display:400,400i|Roboto

### Added Playfair Display Italic
This should not require additional action.

### Removed Playfair Display SC
As far as I can tell, navigation is the only place this is used, so it can be replaced with Playfair Display 

### Removed Open Sans
Correct me if I'm wrong but I don't think this is used anywhere. If it is, please switch that text to Roboto.

### LINKED ISSUE #132: Tutorial "seen already" flag not saved
## Issue Report
- This follows PR #115 
- Basically, when a logged in user closes the Tutorial popup, the "I've seen this tutorial already" flag isn't properly saved into Panoptes.
- As a result, the logged in user will always see the Tutorial popup when she/he visits the /classify page, the same as a non-logged-in user.

### Status
Probably an issue to be solved within the ZRC library. The issue is relatively minor and doesn't meet the requirements for being part of the Beta MVP, but the issue should be fixed soon-ish for the sake of user convenience.

wgranger:
This PR simply closes a few issues that were present. The two main issues are the use of fonts and fixing the tutorial/preferences bug.

Closes #127 
Closes #132

### DIFF
--- a/src/ducks/project.js
+++ b/src/ducks/project.js
@@ -96,8 +96,6 @@ const fetchProject = (id = config.zooniverseLinks.projectId) => {
         type: FETCH_PROJECT_SUCCESS,
         data: project,
       });
-      // dispatch(fetchPreferences(project));
-
     })
     .catch((err) => {
 
--- a/src/ducks/project.js
+++ b/src/ducks/project.js
@@ -96,7 +96,7 @@ const fetchProject = (id = config.zooniverseLinks.projectId) => {
         type: FETCH_PROJECT_SUCCESS,
         data: project,
       });
-      dispatch(fetchPreferences(project));
+      // dispatch(fetchPreferences(project));
 
     })
     .catch((err) => {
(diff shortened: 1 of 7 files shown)


src/ducks/project.js
```diff
@@ -96,7 +96,7 @@ const fetchProject = (id = config.zooniverseLinks.projectId) => {
         type: FETCH_PROJECT_SUCCESS,
         data: project,
       });
-      dispatch(fetchPreferences(project));
+      // dispatch(fetchPreferences(project));
```
shaunanoordin: Unused commented-out code, but this can be cleaned up later.

PR #141, opened 2017-11-02
Use Config to Determine Split Slug

wgranger:
Since we are beginning seven-ten testing on production, we needed to update the split duck to find the correct slug depending on the environment.

### DIFF
--- a/src/ducks/splits.js
+++ b/src/ducks/splits.js
@@ -70,7 +70,7 @@ const fetchSplit = (user) => {
       type: FETCH_SPLIT,
     });
 
-    Split.load(config.projectSlug).then((splits) => {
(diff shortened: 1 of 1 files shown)


src/ducks/splits.js
```diff
@@ -69,7 +70,7 @@ const fetchSplit = (user) => {
       type: FETCH_SPLIT,
     });
 
-    Split.load("wgranger-test/anti-slavery-testing").then((splits) => {
+    Split.load(config.projectSlug).then((splits) => {
```
shaunanoordin: Wait, this should be `config.zooniverseLinks.projectSlug` . Let me post a new update commit.

PR #145, opened 2017-11-03
Fix Logic Issues From Internal Review

wgranger:
Two issues are fixed with this PR:

1) Text modifiers account for [unclear] being a word that can wrap correctly with other text modifiers.
2) On grabbing a new subject, the frame reverts back to 0.

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -116,7 +116,7 @@ class SelectedAnnotation extends React.Component {
     const padEnd = new RegExp(endReg, 'ig');
     const unclearPad = new RegExp(unclearReg, 'ig');
 
-    const paddedText = value.replace(padStart, ` [${textTag}]`).replace(padEnd, `[/${textTag}] `).replace(unclearPad, ' [unclear] ').replace(/\s+/g, ' ');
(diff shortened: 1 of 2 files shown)


shaunanoordin:
## PR Update
This just in:
- There was a missing changeFrame() in fetchQueue(). i.e. when we're fetching a new Subject from a queue, the page doesn't reset back to page 1. This fixes Issue 1.
- The word count now properly updates when we insert metatags. Simple fix, actually. This fixes issue 2.
- A fairly insane amount of RegEx logic-checking has been introduced (in the form of cleanText()) to ensure that the use of spaces (especially around [tag]s ) are legit, to ensure our word count is legit.
  - This means the app has an auto-correct function that cleans up [tags] and `"   extra   spaces 
  "`, but it ONLY triggers when we add new tags.
  - I'm hesitant to add the auto-correct function to whenever the user starts typing - it's quite easy to do so (i.e. hook cleanText() to onTextUpdate() ) but it may be a terrible idea to mess with standard user input.

### Status
@wgranger Since I've added some very noticeable updates, I'm returning review duties to you.

wgranger:
LGTM, thanks for catching these issues with your initial review. Approving for merge. Also, I agree that I don't think the auto-correct function should fire each time the user types. I think that's a lot of unnecessary scripts run and it would confuse the user to see text jump around while typing.

PR #151, opened 2017-11-06
Show Gray Lines on Individual Split

### LINKED ISSUE #146: Individual Users Should See Grayed Out (Retired) Lines
This is something that passed under the radar in planning. It was my assumption that users in the individual group would _never_ see previous annotations from Caesar's reducer. However, users in the individual group should see a line that has been previously annotated *only* if that individual line has reached retirement and is grayed out.

wgranger:
This PR addresses two issues:

1) Show gray consensus lines for both collaborative *and* individual splits. Closes #146 
2) Adjust the toggle state for "toggle prev. marks." Now, the user may toggle between seeing all/previous annotations, seeing only user annotations, seeing only current annotation. *Note:* a check was made to only toggle between all marks and no marks if no previous annotations are present.

### DIFF
--- a/src/ducks/subject-viewer.js
+++ b/src/ducks/subject-viewer.js
@@ -31,7 +31,7 @@ const initialState = {
   frame: 0,
   rotation: 0,
   scaling: 1,
-  shownMarks: 0,
+  shownMarks: MARKS_STATE.ALL,
   translationX: 0,
   translationY: 0,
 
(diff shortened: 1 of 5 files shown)


src/ducks/subject-viewer.js
```diff
@@ -184,9 +190,20 @@ const resetView = () => {
...
 
 const togglePreviousMarks = () => {
-  return (dispatch) => {
+  return (dispatch, getState) => {
+    const annotations = getState().previousAnnotations.marks;
+    const variant = getState().splits.variant;
+    const consensusAnnotation = annotations.some(annotation => { return annotation.consensusReached; })
+    const previousAnnotationsPresent = consensusAnnotation || (annotations.length && variant === VARIANT_TYPES.COLLABORATIVE);
+
+    let shownMarks = (getState().subjectViewer.shownMarks + 1) % 3;
```
shaunanoordin: Nice use of enumerated values! Ha, now I understand why you're using numbers instead of strings to represent each state.

One suggestion to improve any enumerated-value-sequencifier code is to do it like...
`... = (getState()subjectViewer.shownMarks + 1) % number_of(MARKS_STATE)`

This is only a minor suggestion though, because creating that number_of() function would probably take more brainspace than necessary, which is a given since JS doesn't have a natural `enum` [type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/enum)

wgranger:
Although minor issues, I addressed the comments above with the most recent commit.

PR #153, opened 2017-11-07
Wait for Splits to Retrieve Workflow

### LINKED ISSUE #148: A Tale of Two Workflows
## Core Change Request
Instead of having only 1 workflow split between our A-split (solo transcription) and B-split (collaborative transcription) users, we now need **two Panoptes workflows,** one for each split.

This is a significant change to the project's planned implementation, so please note some...

### Very Important Notes
1. **Each Panoptes workflow is an exact mirror of the other.** Same subjects, same tasks, etc, just different users.
2. As of the latest discussion on the ZooSlack's IMLS channel, the A-B Split will only **run for a month or two** before we finally choose to go all in with either solo transcription or collaborative transcription.

The rationale for this mostly comes down to the scientific data; the research team wants to be able to much more easily separate the data input coming in from Split A and Split B.

### Implementation Notes
A few things to be aware of in implementing this.

**PFE:**
- [ ] The A-split and B-split workflows need to match each other, _including the task label._ Be super aware that the first Annotation task should be **T1, not T0.**
- [ ] The act of replicating the workflows needs to be replicated on both production and staging.

**Coleman & Caesar:**
- [ ] Coleman needs to be informed of the new workflow IDs.
 
**Front-End Coding:**
- [ ] Don't forget the config.js file for both staging and production.
- [ ] [MAJOR] The lifecycle for the app needs to be changed; notably, the Seven-Ten A/B split needs to be completed _before_ the app can be displayed to the user. 
  - Very very specifically, the lifecycle on the app loading event needs to be like this...
```
1. Show "Loading app..." status
2. Load the Project data
3. Get logged in user details (if applicable)
4. (requires 3's success) Select the A-B split from Seven-Ten (failsafe required in case Seven Ten is dead)
5. (requires 4's success, or failsafe) Load the Workflow data 
6. (requires 2 and 5's successes) Hide "Loading app..." status, show the full app
```

### Status
This is now a major issue that needs to be tackled for the ASM Beta.

wgranger:
Closes #148 

This PR successfully calls a collaborative workflow based on which split group a user is placed into. However, there is still a lingering issue of resolving the first task which can be seen in the issue above.

**Note**: This has been tagged WIP until the task number issue above has been solved.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -4,7 +4,6 @@ import { connect } from 'react-redux';
 import { ZooFooter } from 'zooniverse-react-components';
 import { fetchProject } from '../ducks/project';
 import { fetchWorkflow } from '../ducks/workflow';
-import { fetchSplit } from '../ducks/splits';
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
 import Dialog from './Dialog';
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -17,12 +17,18 @@ import { SPLIT_STATUS } from '../ducks/splits';
 import GALogAdapter from '../lib/ga-log-adapter';
 import GoogleLogger from '../lib/GoogleLogger';
 import GeordiLogAdapter from '../lib/geordi-log-adapter';
+import { checkLoginUser } from '../ducks/login';
+
 
 class App extends React.Component {
   constructor(props) {
     super(props);
 
     this.googleLogger = null;
+
+    if (!props.initialised) {
+      props.dispatch(checkLoginUser());
+    }
   }
 
   returnSomething(something) { // eslint-disable-line class-methods-use-this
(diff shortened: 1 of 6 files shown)


src/components/App.jsx
```diff
@@ -17,12 +17,18 @@ import { SPLIT_STATUS } from '../ducks/splits';
...
 
 class App extends React.Component {
   constructor(props) {
     super(props);
 
     this.googleLogger = null;
+
+    if (!props.initialised) {
+      props.dispatch(checkLoginUser());
+    }
```
wgranger: I moved this to App without realizing later that I don't think it's necessary. I'm open to move it back if we decide to do so.

shaunanoordin: Although I still need to test this, moving `checkLoginUser()` to the root `App` component _makes sense_ if we want user state to be a root-level concern. (We do.)

IIRC, the reason that `checkLoginUser()` originally sat in `Header` was because it was part of zoo-reduxify's `AuthHeader` component, which handled the user state logistics in a single, easy to grok component.

shaunanoordin:
To confirm: tests look good on localhost, for a logged in user with a B-split, and a non-logged in user (who defaults to the A-split).

PR #154, opened 2017-11-07
Hide Secret Metadata Columns

### LINKED ISSUE #152: Hide Metadata Info.
Pointing to https://www.zooniverse.org/help, we usually hide header info preceded with a pound sign (#). However, this isn't the current implementation when viewing subject info on the classifier. We need to hide information from public view before beta launch.
 
screen shot 2017-11-07 at 9 15 40 am

wgranger:
A quick PR to close #152


wgranger:
Ahh good catch there. I missed the other special characters since we only told the research team about the pound. Yes, we should stick to standards. Look for a commit shortly.

shaunanoordin:
Nice work with the update! 👍  This PR now LGTM, so let's get it rolling.

PR #156, opened 2017-11-07
Restyle Word Count Indicator

### LINKED ISSUE #149: "X out of Y words typed" message needs to be more friendly
## UI/UX Improvement Needed
As @beckyrother pointed out, the glaring red "X out of Y words typed" message looks rather hostile to users, especially when a user first starts transcribing. (It's 0 out of Y words, and loos like an error message is occurring before the user even has a chance to start.)
screen shot 2017-11-06 at 19 02 09

### Recommended Changes
- For starters, ditch the red (i.e. negative feedback) visuals.
- Instead, use the following feedback/messaging system:
  - [ ] If words !== dots-1, display _"X out of Y words typed"_ in _neutral colours._
  - [ ] If words === dots-1, display _"Y out of Y words typed 👍 "_ in _approval/good colours,_ and maybe with a 👍  icon to indicate they've done everything that's expected.
  - [ ] If input.contains('[tags]'), display _"X out of Y words typed. However, we noticed you added insertion/deletion/unclear tags, so don't worry too much about the word count. If you want to know more about how to handle weird insertion/deletion cases, [click here](https://giphy.com/gifs/idk-shrug-power-rangers-y65VoOlimZaus) "_
- Allow users to submit data EVEN if words don't match number of dots. Observe the following possibilities:
  - [ ] When the user hits Done/Done & Talk, AND words === dots-1, proceed normally.
  - [ ] When the user hits Done/Done & Talk, AND words !== dots-1, prompt the user with a confirmation dialogue: _"We expected you to type Y words since you marked Y+1 dots, but you typed in X words. Are you sure you're done? Yes/No"_
  - [ ] When the user hits Done/Done & Talk, AND words !== dots-1 AND [tags] detected, prompt the user with a confirmation dialogue: _"[insertion][deletion] can be pretty tricky and can mess up the expected word count. You can check [our guide](https://giphy.com/gifs/idk-shrug-power-rangers-y65VoOlimZaus) if you're unsure, but if you're confident, we can proceed. Done? Yes/No"_

### Status
Priority TBC, and the text copy needs some attention, but I'd like to see this running for Beta since it ties in with the other "what to do with [insertion] [deletion]" issue/implementation discussion.

wgranger:
Currently, the sole commit in this PR handles the restyling mentioned in #149. However, I'm still unclear about the following:

- Are we allowing users to submit a classification if the word count doesn't match the number of dots (using a disclaimer)?

Closes #149

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -110,17 +110,17 @@ class SelectedAnnotation extends React.Component {
       annotationText: this.cleanText(value),
     });
   }
-  
+
   /*  cleanText() makes the user's transcription text more palatable to the
       aggregation engine.
-      
+
       NOTE: cleanText() can be added to onTextUpdate() to create an
       "auto-correct as you type" feature, but boy oh boy we need to be careful
       about messing around with standard user input. (@shaun 20171107)
    */
   cleanText(text) {
     return text
-  
+
     //Generic tags
     .replace(/(\[\w+\])/g, ' $1')  //When  we see [tag], add a space in front of it.
     .replace(/(\[\/\w+\])/g, ' $1')  //When  we see [/tag], add a space after it.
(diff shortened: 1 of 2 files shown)


shaunanoordin:
> Are we allowing users to submit a classification if the word count doesn't match the number of dots (using a disclaimer)?

Good point, we should _not_ allow this. The simplest thing to do is to make sure that when a user Cancels, the Annotation is deleted _if_ the user hasn't typed in any words yet.

Wait, let me take a look at a the code, I think we can have a very straightforward solution. Something like...

```
onCancel() {
  if (annotation.words.trim().length() === 0 && textInput.value.trim().length()) {
    this.deleteAndClose();
  } else {
    this.close();  //As normal
  }
}
```

beckyrother:
Looks great, thanks!

PR #158, opened 2017-11-08
[RFC] Only Use Beta Subject Sets

wgranger:
This is necessary for beta launch. Instead of giving users the option to choose which subject set they draw from, the beta subject set will be the only one used. This should be removed eventually for full launch.

Currently RFC because there might be the same issue with the `subject_set_id` query failing to pull subjects from the correct set, but I'd like another set of eyes on the code to confirm.


shaunanoordin:
Update: resolved conflicts in config.js. Basically, the config now contains data for both betaSubjectSet and the collabWorkflowId

PR #159, opened 2017-11-08
Cancelling a new Annotation now deletes it

shaunanoordin:
## PR Overview
- With this PR, when a user creates a new Annotation and then proceeds to Cancel it (or clicks on the X to close the window) on the Transcription popup, that new Annotation is _deleted._
  - For our purposes, a 'new Annotation' can be implied when the selected annotation has zero transcribed text.
  - This prevents users from submitting Annotations with no text, simply by cancelling out of a newly created Annotation (which has no text).
- Clicking Cancel or close on an existing Annotation simply reverts the Annotation to its previous state, (i.e. before the Transcription popup appeared,) as normal.

### Status
Ready for review, @wgranger

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -235,7 +235,23 @@ class SelectedAnnotation extends React.Component {
   }
 
   cancelAnnotation() {
-    if (this.props.onClose) { this.props.onClose() };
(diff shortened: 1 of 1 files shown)


wgranger:
LGTM. Good work doing a lot with a few lines of code.

PR #160, opened 2017-11-09
Successfully fetching Subjects now properly resets Annotations, Previous Annotations, and Classifications

shaunanoordin:
## PR Overview
The original issue that this PR attempts to fix is phrased this way: _When a user makes Annotations on Subject A, then navigates away from the Classifier page, then returns, they will see Subject B loading, but with Subject A's Annotations._

This was due to the fact that the Subject fetching lifecycle was somewhat broken. This PR improves the lifecycle by _ensuring_ that when a new Subject is successfully fetch, all items that _should follow that event in the lifecycle_ will accordingly be reset into a blank slate.

i.e., when a new Subject is fetched successfully,
- reset Annotations
- reset Previous Annotations
- create a new, empty Classification
- reset the Subject Viewer so it's pointing back to page 1 of the Subject letter.

### Things To Keep An Eye Out For
- Are Previous Annotations still displaying properly for a B-split user?
- If a user moves away from the Classifier page, they will lose all the work they've done. This can be SUPER ANNOYING.

### Status
WIP - due to the change in the Subject fetch lifecycle, I'm still trying to confirm that the B-split works.

I'm thinking that, ultimately, `fetchAnnotations()` in `SubjectViewer.componentWillReceiveProps` needs to be folded into the lifecycle instead of being something monitored by a component.

Also, maybe have a "ClassifierPage.onUserLeavesPage = confirmUserWantsToDoThis"

### DIFF
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -547,7 +547,7 @@ SubjectViewer.defaultProps = {
   annotationInProgress: null,
   annotations: [],
   //--------
-  selectedAnnotations: null,
+  selectedAnnotation: null,
 };
 
 SubjectViewer.contextTypes = {
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -192,10 +192,18 @@ class SubjectViewer extends React.Component {
     //Make sure we monitor visible size of Subject Viewer.
     window.addEventListener('resize', this.updateSize);
     this.updateSize();
-    this.props.dispatch(fetchSubject());  //Fetch the first subject. This will also ensure a clean slate for Annotations, Previous Annotations, and Classifications.
+    
+    //Fetch the first subject, IF no subject has yet been loaded.
+    //Fetching a subject will also ensure a clean slate for Annotations,
+    //Previous Annotations, and Classifications.
+    this.props.dispatch(fetchSubject(
+      undefined,  //JS Quirk: this ensures we're fetching the default value for the 'id' parameter (as defined in fetchSubject())
+      true,  //Initial fetch only, meaning ignore this action call if a Subject is already being fetch/has already been loaded.
+    ));
   }
 
   componentWillReceiveProps(next) {
+    //An annotation was just selected.
     if (!this.props.selectedAnnotation && next.selectedAnnotation) {
       this.setState({
         annotation: <SelectedAnnotation annotation={next.selectedAnnotation} onClose={this.closeAnnotation} />
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -26,7 +26,7 @@ import SVGImage from '../components/SVGImage';
 import AnnotationsPane from '../components/AnnotationsPane';
 import ZoomTools from '../components/ZoomTools';
 import { Utility } from '../lib/Utility';
-import { fetchSubject, setImageMetadata, SUBJECT_STATUS } from '../ducks/subject';
+import { fetchSubject, setImageMetadata } from '../ducks/subject';
 import { getSubjectLocation } from '../lib/get-subject-location';
 import SelectedAnnotation from '../components/SelectedAnnotation';
 
@@ -473,7 +473,6 @@ SubjectViewer.propTypes = {
   currentSubject: PropTypes.shape({
     src: PropTypes.string,
   }),
-  subjectStatus: PropTypes.string,
   //--------
   contrast: PropTypes.bool,
   frame: PropTypes.number,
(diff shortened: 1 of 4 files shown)


shaunanoordin:
## PR Update
- This PR now cleans up the fetch Subject lifecycle even further.
- fetchPreviousAnnotations is now handled by the fetchSubject() action, instead of being handled by the SubjectViewer component.

### Status
@wgranger this PR is now out of WIP and ready for review.

Tested on A-split and B-splits on localhost, with the B-splits showing Previous Annotations prpoerly; thanks to Will for arranging the proper settings on Panoptes-staging.

My only concern is that a user who navigates out of the page will find it super annoying that all the work she or he had done is gone. I'm investigating that "Are you sure you wanna navigate away from this page?" hook, but that I may put that in a separate PR.

wgranger:
Concerning your "Are you sure you want to navigate away..." I should also have a PR up soon that allows users to save annotations, so this should help alleviate the burden of losing annotations.

wgranger:
This PR cleans up code a good bit, so I'm a fan of the readability improvement. 

However, some tweaks are needed with this code because I'm still seeing an issue of annotations persisting in the classifier when navigating through the site. Ex. if I make some annotations on the classifier, go to the home page, and then revisit the classifier, all my annotations are reset.

I'm hesitant to merge this PR until the issue above is resolved.

src/containers/SubjectViewer.jsx
```diff
@@ -193,7 +192,7 @@ class SubjectViewer extends React.Component {
     //Make sure we monitor visible size of Subject Viewer.
     window.addEventListener('resize', this.updateSize);
     this.updateSize();
-    this.props.dispatch(fetchSubject());  //Fetch the first subject.
+    this.props.dispatch(fetchSubject());  //Fetch the first subject. This will also ensure a clean slate for Annotations, Previous Annotations, and Classifications.
```
wgranger: This dispatch call now clears annotations made on the classifier when navigating to another location on the app and returning to the classifier (ex. classifier -> About -> classifier erases work that was done).

shaunanoordin:
> Ex. if I make some annotations on the classifier, go to the home page, and then revisit the classifier, all my annotations are reset.

That's correct; all annotations are wiped whenever a new Subject loads, and it just so happens that when a user visits/revisits the Classifier page, it will _always_ trigger a new Subject load. ( `SubjectViewer.componentDidMount() = { fetchSubject(); }` )

There are two solutions to this:
1. Make sure the Classifier page (SubjectViewer component, actually) _only_ loads a new subject if one hasn't already been loaded, OR,
2. When the user attempts to leave the Classifier page, check with the user if they're sure they want to do so.

Option 1 proved tricky due to a quirk with the React lifecycle - I want to trigger fetchSubject() on componentDidMount _if_ this.props.currentSubject is null (or more accurately, `subjectStatus === IDLE`, but at that stage (i.e. before receiving the store values via componentWillReceiveProps), this.props.currentSubject _is always null._

I was looking at Option 2, as noted in my comment:
> My only concern is that a user who navigates out of the page will find it super annoying that all the work she or he had done is gone. I'm investigating that "Are you sure you wanna navigate away from this page?" hook, but that I may put that in a separate PR

However, if you're not comfortable with merging until the matter is addressed, no worries, I'll look into implementing the fix in this PR.

shaunanoordin:
## PR Update
- The props for the SubjectViewer component has been cleaned up - hopefully it's easier to see what's happening in the code now.
- I've gone with Option 1 as noted above:
  - The initial fetchSubject() in SubjectViewer now indicates that it's an "initial fetch only", and
  - fetchSubject() now has an optional flag to indicate it's an "initial fetch only", indicating that the action should be skipped if a Subject already exists.

### Status
Ready for a re-review, @wgranger

src/containers/SubjectViewer.jsx
```diff
@@ -457,14 +457,18 @@ class SubjectViewer extends React.Component {
...
   //--------
+  splits: PropTypes.object,
+  user: PropTypes.shape({
+    id: PropTypes.string
+  }),
+  //--------
   currentSubject: PropTypes.shape({
     src: PropTypes.string,
   }),
+  subjectStatus: PropTypes.string,
```
wgranger: Is this prop being used by the SubjectViewer?

shaunanoordin: Wait, derp, I moved that logic to fetchSubject() already.

src/containers/SubjectViewer.jsx
```diff
@@ -529,11 +535,13 @@ SubjectViewer.defaultProps = {
...
   //--------
+  previousAnnotations: [],
+  //--------
   annotationsStatus: ANNOTATION_STATUS.IDLE,
   annotationInProgress: null,
   annotations: [],
-  splits: null,
-  user: null
+  //--------
+  selectedAnnotations: null,
```
wgranger: This should not be pluralized.

shaunanoordin: Ah! Thanks for catching this.

wgranger:
Thanks for making those fixes. LGTM. Merging.

PR #162, opened 2017-11-10
Open Gold Standard Mode

wgranger:
This PR opens a Gold Standard Mode which will have access to a separate workflow to record gold standard annotations. I initially set out to refactor the code a bit, but I feel like this is a case where a good bit of code is needed to accomplish a simple task (fetching roles, running conditionals, rendering based on these results).

Currently, only an owner and expert can enter gold standard mode via the home screen. I came across two minor issues with this branch, but I'm not sure it's worth a fix since (as to my knowledge) we only have one person currently who will be annotating under this mode.

1) If a user logs out while in gold standard mode, the mode will not be reset.
2) There is a reloading and jumping of the home page when gold standard mode is toggled (probably because a new workflow is fetched).

If we were to have many experts classifying, this might be worth more attention, but I'm not sure it's necessary given such a low number of experts. However, I'm happy to push a fix if you find this blocking.

### DIFF
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -89,7 +89,7 @@ class Home extends React.Component {
   }
 
   render() {
-    const isExpert = this.props.userRoles.indexOf(EXPERT || OWNER) >= 0;
+    const isExpert = this.props.userRoles.some(role => { return role === EXPERT || role === OWNER });
 
     return (
       <main className="app-content home-page">
@@ -119,7 +119,7 @@ class Home extends React.Component {
           </div>
           */}
           {isExpert && (
-            <div>
+            <div className="gold-standard-toggle">
               <input type="checkbox" onClick={this.toggleGoldStandard} id="goldStandard" checked={this.props.goldStandardMode} />
               <label htmlFor="goldStandard">Gold Standard Mode</label>
             </div>
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -8,11 +8,14 @@ import { ZooniverseLogotype, ZooniverseLogo } from 'zooniverse-react-components'
 import { config, subjectSets } from '../config';
 import SocialSection from '../components/SocialSection';
 import { selectSubjectSet } from '../ducks/subject';
+import { setGoldStandard } from '../ducks/workflow';
 import Divider from '../images/img_divider.png';
 import BostonLogo from '../images/BPL_logo.jpg';
 import IMLSLogo from '../images/imls_logo.png';
 
 const BUFFER = 50;
+const EXPERT = 'expert';
+const OWNER = 'owner';
 
 class Home extends React.Component {
   constructor() {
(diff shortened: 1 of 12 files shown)


src/components/Home.jsx
```diff
@@ -8,11 +8,14 @@ import { ZooniverseLogotype, ZooniverseLogo } from 'zooniverse-react-components'
 import { config, subjectSets } from '../config';
 import SocialSection from '../components/SocialSection';
 import { selectSubjectSet } from '../ducks/subject';
+import { setGoldStandard } from '../ducks/workflow';
 import Divider from '../images/img_divider.png';
 import BostonLogo from '../images/BPL_logo.jpg';
 import IMLSLogo from '../images/imls_logo.png';
 
 const BUFFER = 50;
+const EXPERT = 'expert';
```
shaunanoordin: [minor] Since "user role" is a sorta-enumerated value, it might make sense to conceptually organise this into `const USER_RULES = { EXPERT: 'expert', OWNER: 'owner' }`

[minor] Would we also need a USER_ROLES.GENERIC type? I need to check the rest of the code.

This would be matters for future improvement, though.

shaunanoordin:
[minor] The "Gold Standard" option on the home page is still present after the user clicks "Log Out",  and only disappears after the page refreshes (or re-renders). That said, this is a minor issues as this only affects our Expert users.

shaunanoordin:
Wait, hang on... does
```
//in ducks/classificaitons.js
classification.gold_standard = true;
```
actually do anything?

Two things:
1. Are we sure it's `gold_standard` and not `metadata.gold_standard` ? (I think it's correct, but I don't have a baseline to check against.)
2. All other Classification data was set by using `classification.update({...}).save()` - is this the 'correct' way to set a value?

Side note: question also applies to
```
//in ducks/classificaitons.js
classification._workflow = getState().workflow.data;
classification._subjects = [getState().subject.currentSubject];
```
but to a lesser degree of importance. It doesn't seem that necessary for the Classification to work, so huh.

wgranger:
`gold_standard` is a standalone tag apart from `metadata`. Using the Apiary as a resource https://panoptes.docs.apiary.io/#reference/classification/classification

wgranger:
Yes, the classification resource needed an `.update` when the `gold_standard` tag was added, so this most recent commit should fix things. Also, there is another bug I've noticed that should be fixed, but correct me if my logic is wrong. Whenever a new subject set is selected, the subject queue should clear, so that a new queue can be loaded from the new subject set. This likely applies to this PR, but I believe it should be fixed in another branch since it applies to any subject set selection.

wgranger:
Ok, fix was actually just made with a recent commit for the following reason.

1) If a user begins transcribing (and is not in GS mode) a queue will be set using a different subject set.
2) If that user goes back to the main page, turns on GS mode, and then enters the classifier, the same queue from earlier will be used (which doesn't contain the GS subject set).
3) In order to fetch the new subject set upon toggle, the queue must be cleared with `subjectStatus` set to idle so the subject viewer and subject duck knows the fetch a new queue.

This same logic should _probably_ be introduced to the app once we turn back on subject selection, so users are getting a fresh queue each time a new subject set is selected.

shaunanoordin:
## PR Review
The latest commits fix the primary issue of the `classification.gold_standard` not being triggered properly. All others forms of standard functionality - e.g. transcribing as a logged in/non-logged in user, and submitting Classifications in non-Gold Standard modes - remain stable.

### Status
LGTM! Merging, go.

PR #164, opened 2017-11-13
Save Classifications for Later

wgranger:
This is the last major bit of code being opened before beta. There is a lot going on here, and I changed the approach to save classifications on Panoptes instead of dumping them onto `localStorage`. I'm open to any suggestions and can explain any bit of code that might be perplexing. 

Note: This is not mandatory for beta but would be nice to introduce in time for the full launch.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -73,7 +73,7 @@ class App extends React.Component {
     const showTitle = path === '/classify';
 
     return (
-      <div>
+      <div style={{ "position": "relative" }}>
         <Header />
         <ProjectHeader showTitle={showTitle} />
         {this.props.children}
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -73,7 +73,7 @@ class App extends React.Component {
     const showTitle = path === '/classify';
 
     return (
-      <div style={{ "position": "relative" }}>
+      <div>
         <Header />
         <ProjectHeader showTitle={showTitle} />
         {this.props.children}
--- a/src/components/ClassificationPrompt.jsx
+++ b/src/components/ClassificationPrompt.jsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import apiClient from 'panoptes-client/lib/api-client';
 import { connect } from 'react-redux';
 import { retrieveClassification } from '../ducks/classifications';
+import { fetchSubject } from '../ducks/subject';
 
 class ClassificationPrompt extends React.Component {
   constructor(props) {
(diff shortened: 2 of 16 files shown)


shaunanoordin:
This PR currently has conflicts; I'll attempt to resolve them via the GitHub editor.

shaunanoordin:
@beckyrother , can you please take a look at my comments above? This PR is functional, but I have concerns about the messaging and I thought it'd be good if you could give my ideas a sanity check.

ALSO: I just added you as a collaborator to this project so now I can @ mention you properly, yaaay! Why didn't I do this earlier and why didn't I copy a proper "project settings" template for this repo

beckyrother:
Thanks @shaunanoordin ! 

The idea for the Save and Close button was to alleviate any anxiety a user might have about losing work when they leave a session before submitting entries. If I'm reading this correctly, is the suggestion to change the text to just Save and let the user leave when they're ready? I think this is a fine option and basically does the same thing – it's just a placebo, right, since we autosave all changes? I guess the only argument there would be if a person just wants to leave without saving changes, it might be confusing to come back to that session and find the work they didn't want to save also there. 

The modal when they come back to the page makes sense and I like the messaging.

Sorry for the stream-of-consciousness; let me know if clarification is needed.

shaunanoordin:
Thanks @beckyrother ! The "Save" option _isn't_ a placebo, as - at the moment - there is no autosave feature. The Save feature for ASM works more like classic Microsoft Word from the 1990s _(where if you forgot to Ctrl+S after 2 hours of writing your thesis, you'd start screaming to the heavens when the power suddenly goes out)_ instead of modern 2010s Google Docs _(where all changes are saved every few seconds, so all your mistakes will be remembered, forever.)_

Since this PR works like the classic "save it or lose it" style as opposed to an autosave style which modern users might incorrectly assume, I thought it's extremely important to message to the users that
1. You really should save your work every now and then to avoid losing everything you did, and
2. But don't you confuse "Save" with "Submit", you hear?

(Hmm... a thought: does "Save Progress" sound obvious?)

NOTE: @wgranger I'm actually OK to merge this right now, then approach the improved messaging I discussed later - _provided we don't deploy to production_ until that's addressed.

beckyrother:
That sounds great @shaunanoordin ! "Save Progress" sounds great. Is there a way to warn people when they leave the page that their classifications haven't been / won't be saved ?

wgranger:
This branch was updated along with our most recent comments. There was discussion in a recent call about adding a _last saved: (time)_ notification, but I'm punting this to a future PR.

shaunanoordin:
## PR Review (Update)
The latest updates have addressed the concerns about how the feature is communicated, and the Save Progress feature is looking better than ever.

As usual, the features test OK (can save progress, and re-load progress on a page refresh) on OSX+Chrome using zootester1, and I'll re-test again once merged to master to address the concern: _"I altered the way annotations are saved just slightly because I was getting an error from the api based on how the branch was set up before the most recent two commits."_

### Status
👍  Awesome! I'm merging this, and deploying after a quick test on master.

shaunanoordin:
ERROR ERROR
I can't submit Classifications now - reverting branch and seeing WTF is up

PR #165, opened 2017-11-13
Fix Modal Placement on IE

### LINKED ISSUE #18: IE11 Compatibility
## Issue Overview
* We currently have not implemented compatibility with Internet Explorer 11.
* There are two parts to consider:
  * Polyfills (e.g. [Polyfill.io](https://polyfill.io/v2/docs/), notably for Promises) should take care of the _functionality_ side of things.
  * As for the presentation side of things, the CSS may require more hacky solutions - but we'll cross those bridges and/or set fire to them when necessary.
* More important question: _do we want IE11 compatibility?_ We already have Edge compatibility (woohoo!) and I'm not sure what the user base for this project will be like, and I generally prefer adding IE11 hacks unless absolutely necessary.

wgranger:
This was a relatively easy fix. The dialog modal wasn't placing right on IE because the browser can't recognize `window.scrollY`. Alternatively, I used `window.pageYOffset`.

I believe this is the final fix needed to close #18 
Closes #18

### DIFF
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -22,7 +22,7 @@ class Dialog extends React.Component {
     const width = 800;
     const height = 425;
     const x = window.innerWidth / 2 - (width / 2);
-    const y = window.innerHeight / 2 - (height / 2) + window.scrollY;
+    const y = window.innerHeight / 2 - (height / 2) + window.pageYOffset;
 
     const defaultPosition = { x, y, height, width };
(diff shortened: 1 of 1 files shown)


shaunanoordin:
## PR Review
This PR fixes the position of of all Dialog popups (e.g. Subject Info) on IE11, which previously 'stuck' to the top of the screen instead of appearing in the centre of the window.

Tested OK on Win7+IE11, Win8+Edge, and OSX+Chrome62/FF56/Safari

### Status
LGTM, merging 👍

PR #168, opened 2017-11-14
Reveal Feedback Forms

wgranger:
This PR adds a link for feedback forms on the project header in preparation for beta launch. A user will be directed to an individual or collaborative form depending on which variant group is used.


shaunanoordin:
The Google Forms now have their permissions set correctly. I will proceed to update the URLs accordingly, brb.

PR #174, opened 2017-11-17
Crib Sheet

### LINKED ISSUE #9: Build the Crib Sheet
At present, some more thought needs to go into exactly how this will function. This crib sheet will display information from past classifications to aid the volunteer in annotations.

wgranger:
This component likely still needs some work, so it will be WIP for now. I still need to dummy check the code, but I'm putting it up now mainly to skim through and also test out. At this point, I mainly want someone to test it out to determine if there is something obviously out of place. Some other questions to consider.

- Should we add a keyboard feature to escape from cropping mode? Ex: Pressing `esc` goes back to navigation mode on the subject viewer.

Closes #9 

TODO: 
- Make the `SaveClip` component's dialog modal more narrow.

### DIFF
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -27,8 +27,8 @@ class Dialog extends React.Component {
   render() {
     const width = this.props.isPrompt ? 450 : 800;
     const height = 425;
-    const x = window.innerWidth / 2 - (width / 2);
-    const y = window.innerHeight / 2 - (height / 2) + window.pageYOffset;
+    const x = (window.innerWidth / 2) - (width / 2);
+    const y = (window.innerHeight / 2) - ((height / 2) + window.pageYOffset);
 
     const defaultPosition = { x, y, height, width };
     const enableResize = this.props.enableResize ? { bottomRight: true } : false;
@@ -50,7 +50,11 @@ class Dialog extends React.Component {
         minWidth={400}
         resizeHandlerClasses={resizeClass}
       >
-        <div className="popup dialog" ref={(c)=>{this.popupBody=c}} onClick={(e) => { return e.target === this.popupBody && this.close(e); }}>
+        <div
+          className="popup dialog"
+          ref={(c) => { this.popupBody = c; }}
+          onClick={(e) => { return e.target === this.popupBody && this.close(e); }}
+        >
           <div className="popup-content dialog-content">
             <div>
               {this.props.title.length ? (
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -45,6 +45,7 @@ class Dialog extends React.Component {
       <Rnd
         default={defaultPosition}
         enableResizing={enableResize}
+        dragHandlerClassName={'.handle'}
         minHeight={400}
         minWidth={400}
         resizeHandlerClasses={resizeClass}
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -37,6 +37,10 @@ class Dialog extends React.Component {
       return React.cloneElement(child, { onClose: this.close });
     });
 
+    const children = React.Children.map(this.props.children, (child) => {
+      return React.cloneElement(child, { onClose: this.close });
+    });
+
     return (
       <Rnd
         default={defaultPosition}
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -37,10 +37,6 @@ class Dialog extends React.Component {
       return React.cloneElement(child, { onClose: this.close });
     });
 
-    const children = React.Children.map(this.props.children, (child) => {
-      return React.cloneElement(child, { onClose: this.close });
-    });
-
     return (
       <Rnd
         default={defaultPosition}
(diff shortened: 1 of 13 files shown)


src/components/ShowMetadata.jsx
```diff
@@ -4,7 +4,7 @@ import Divider from '../images/img_divider.png';
 
 const ShowMetadata = ({ metadata }) => {
   return (
-    <div className="show-metadata">
+    <div className="handle show-metadata">
```
wgranger: The `.handle` class was added to several components in order to avoid the "drag while typing" bug that appeared with the selected annotation input box. In this case, all children of a `Dialog` component are given the `.handle` class if they need to drag, where the `CribEdit` component toggles this on and off depending on if the input box is selected. 

Without the introduction of this `dragHandlerClassName` (see `Dialog` component), any clickable space in a `Dialog` component (including children) would initiate a drag.

wgranger:
This has been cleaned up using a linter. You might see a couple corrections to lines not affected by this PR, and that is satisfying the unruly linter.

src/components/Dialog.jsx
```diff
@@ -18,59 +18,69 @@ class Dialog extends React.Component {
...
+    return this.stopEvent(e);
+  }
+
   render() {
     const width = 800;
     const height = 425;
-    const x = window.innerWidth / 2 - (width / 2);
-    const y = window.innerHeight / 2 - (height / 2) + window.pageYOffset;
+    const x = (window.innerWidth / 2) - (width / 2);
+    const y = (window.innerHeight / 2) - ((height / 2) + window.pageYOffset);
```
shaunanoordin: Why is `window.pageYOffset` set within the brackets? This will cause the maths to expand so that it becomes `window.innerHeight / 2 - height / 2 MINUS window.pageYOffset`, so instead of having the dialog more-or-less anchored to the centre of the browser window, _it causes the dialog to appear higher in the page the lower down you've scrolled._

shaunanoordin:
## Quick thoughts
- Cribsheet's "snapshotting" function works, nice!
- Cribsheet's "view my catalogue" works, but is very rudimentary - for example, what if I snapshot a large square? Or a very long rectangle? The cribsheet's view takes a thumbnail of my snapshot, and displays only a square - are there plans for allowing me to view my full snapshot at full size?
- There is a mathematical problem in `Dialog.jsx` that causes the Dialog to appear higher up the page, the farther down you scroll. (Instead of appearing at the centre of the browser window.)

wgranger:
Most recent push adds:
- Changes text description on cribsheet
- Small snippet (less than 10px dimension) will not save
- Don't allow resizing cribsheet (to keep clean grid view of snippets)
- `SaveClip` component div is narrower
- When in snippet/crop mode, you can escape by hitting `esc`
- On cribsheet, the delete button only appears on snippet hover

TODO:
- Visual indicator of cropping mode
- Transparent background of cribsheet div

shaunanoordin:
@wgranger I detected a small bug that can occur.
- If a user has entered "snapshot"/"clip-saving" mode, but has _not started_ drawing a rectangle - pressing Esc will cancel out of the snapshot mode. 👌 
- If a user has entered "snapshot" mode, and has _started_ drawing a rectangle but _not finished_ - the app goes into a limbo state where it doesn't cancel out of snapshot mode properly. 👎 

screen shot 2017-11-21 at 15 20 14
_You got my annotations in my cribsheeting! You got my cribsheeting in my annotations!_ 

Possible solution:
```
  escapeCrop(e) {
    if (e.keyCode === 27 && this.props.viewerState === SUBJECTVIEWER_STATE.CROPPING) {
      this.props.dispatch(setViewerState(SUBJECTVIEWER_STATE.NAVIGATING));
      //Something needs to be placed here that kicks off a... say... "if a snapshot is in progress, cancel it, yo" call.
    }
  }
```

src/containers/SubjectViewer.jsx
```diff
@@ -243,6 +263,12 @@ class SubjectViewer extends React.Component {
     this.props.dispatch(updateViewerSize(svgW, svgH));
   }
 
+  escapeCrop(e) {
+    if (e.keyCode === 27 && this.props.viewerState === SUBJECTVIEWER_STATE.CROPPING) {
```
shaunanoordin: Can you please change `e.keyCode === 27` to `getKeyCode(e) === KEY_CODES.ESCAPE` ? (See `lib/Utility.js`)

Two reasons:
1. Using KEY_CODES.ESCAPE is somewhat more readable that simply 27.
2. [e.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) is being deprecated, so it's good to future-proof our use of it.

Good news is, we already have a library for understanding keyboard input. Finally, all those hackdays spent refining video game controls have paid off!

(Side note: I think I saw an e.keyCode being used somewhere else too, in an earlier PR - I should check if I remembered to fix that.)

wgranger:
Ok. Some adjustments were made to this branch in response to comments. The only outstanding TODO is to make the cribsheet transparent, but I'm going to wait until #164 is merged, or save this for another PR, since the changes I make will potentially affect any component using the `Dialog` component, and the save classification PR uses several of these.

wgranger:
This has been rebased. I'm going to make the opacity change in a future PR since the dialog box is going to be adjusted a bit. Because of that, I'm taking off the WIP notation.

wgranger:
This was recently rebased and updated to reflect the style changes from #187. There was also a bad rebase that I had to address in the process.

snblickhan:
This would be *nice* for beta 2, but I'm okay with pushing it to full launch if there isn't time.

PR #175, opened 2017-11-20
Introduce Classification Queue

### LINKED ISSUE #173: Long Classifications Fail
There appears to be an issue with classifications failing to submit if a user is working on transcribing for a while (around an hour). At this point, I have been unable to replicate this issue, but we should have some sort of session storage for a classification implemented in the case of a failed classification. In the instance of a failed classification, the classification would be kept in a queue while the token is refreshed and classification submission is attempted again. [This is used in PFE](https://github.com/zooniverse/Panoptes-Front-End/blob/6a228205f1aa258c76fb0d5e79459c246f6abb24/app/pages/project/classify.cjsx#L274).

wgranger:
This PR introduces a classification queue similar to the one used in PFE. It would be good to have a second set of eyes on this functionality. There's a good bit going on in the classifier, and although the code _appears_ sound, it would be nice to get another set of eyes on it.

Also, I was unable to recreate the original bug this solved to fix. I attempted to recreate the bug by beginning a classification and attempting to submit said classification after an hour, so I'm not positive this PR will fix the issue we were experiencing.

Closes #173

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -305,7 +305,7 @@ const submitClassification = () => {
       'metadata.subject_dimensions': subject_dimensions || [],
     });
     queueClassification(classification, user);
-    saveAllQueuedClassifications(dispatch);
+    saveAllQueuedClassifications(dispatch, getState);
   };
 };
 
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -131,8 +131,7 @@ const createClassification = () => {
     - Once ALL the items are processed, we are now ready to prepare our data
       for the next Subject to be fetched.
  */
-const saveAllQueuedClassifications = (dispatch, getState) => {
-  const user = getState().login.user;
+const saveAllQueuedClassifications = (dispatch, user = null) => {
   const QUEUE_NAME = (user)
     ? user.id + '.' + CLASSIFICATIONS_QUEUE_NAME
     : '_.' + CLASSIFICATIONS_QUEUE_NAME;
@@ -304,7 +303,7 @@ const submitClassification = () => {
       'metadata.subject_dimensions': subject_dimensions || [],
     });
     queueClassification(classification, user);
-    saveAllQueuedClassifications(dispatch, getState);
+    saveAllQueuedClassifications(dispatch, user);
   };
 };
 
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -14,8 +14,6 @@ const CLASSIFICATIONS_QUEUE_NAME = 'classificationsQueue';
 
 //Action Types
 const SUBMIT_CLASSIFICATION = 'SUBMIT_CLASSIFICATION';
-//const SUBMIT_CLASSIFICATION_SUCCESS = 'SUBMIT_CLASSIFICATION_SUCCESS';
-//const SUBMIT_CLASSIFICATION_ERROR = 'SUBMIT_CLASSIFICATION_ERROR';
 const SUBMIT_CLASSIFICATION_FINISHED = 'SUBMIT_CLASSIFICATION_FINISHED';
 const CREATE_CLASSIFICATION = 'CREATE_CLASSIFICATION';
 const CREATE_CLASSIFICATION_ERROR = 'CREATE_CLASSIFICATION_ERROR';
@@ -207,6 +205,7 @@ const saveAllQueuedClassifications = (dispatch, getState) => {
           localStorage.setItem(QUEUE_NAME, JSON.stringify(newQueue));
           
           //All done, get the next Subject!
+          dispatch({ type: SUBMIT_CLASSIFICATION_FINISHED });
           dispatch(fetchSubject());  //Note: fetching a Subject will also reset Annotations, reset Previous Annotations, and create an empty Classification.
           dispatch(resetView());
         }
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -125,52 +125,54 @@ const createClassification = () => {
 
 const saveAllQueuedClassifications = (dispatch) => {
   const queue = JSON.parse(localStorage.getItem(FAILED_CLASSIFICATION_QUEUE_NAME));
+  let shownAlert = false;
+
   if (queue && queue.length !== 0) {
     console.log('Saving queued classifications:', queue.length);
+    const newQueue = [];
     queue.forEach((classificationData) => {
       apiClient.type('classifications').create(classificationData).save()
         .then((actualClassification) => {
           console.log('Saved classification', actualClassification.id);
 
-          try {  //Fix: IE11 doesn't know what to do with Split.classificationCreated()
+          try {
             Split.classificationCreated(actualClassification);
           } catch (err) { console.error('Split.classificationCreated() error: ', err); }
 
           actualClassification.destroy();
-          const indexInQueue = queue.indexOf(classificationData);
-          queue.splice(indexInQueue, 1);
+
           try {
             localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(queue));
             console.info('Saved a queued classification, remaining:', queue.length);
-            console.info('ducks/classifications.js submitClassification() success');
             dispatch({ type: SUBMIT_CLASSIFICATION_SUCCESS });
           } catch (error) {
             console.error('Failed to update classification queue:', error);
           }
           const { workflow, subjects } = actualClassification.links;
           dispatch(addAlreadySeen(workflow, subjects));
-
         })
-        .catch((error) => {
-          console.error('Failed to save a queued classification:', error);
+        .catch((err) => {
           console.error('ducks/classifications.js submitClassification() error: ', err);
           Rollbar && Rollbar.error &&
           Rollbar.error('ducks/classifications.js submitClassification() error: ', err);
-          alert('ERROR: Could not submit Classification');  //TODO: better messaging
           dispatch({ type: SUBMIT_CLASSIFICATION_ERROR });
-          switch (error.status) {
+
+          if (!shownAlert) {
+            alert('ERROR: Could not submit Classification');  //TODO: better messaging
+            shownAlert = true;
+          }
+
+          switch (err.status) {
             case 422: {
-              const indexInQueue = queue.indexOf(classificationData);
-              queue.splice(indexInQueue, 1);
-              try {
-                localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(queue));
-              } catch (err) {
-                console.error('Failed to update classification queue:', err);
-              }
               break;
             }
             default:
-              return null;
+              try {
+                newQueue.push(classificationData);
+                localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(newQueue));
+              } catch (error) {
+                console.error('Failed to update classification queue:', error);
+              }
           }
         });
     });
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -144,9 +144,6 @@ const saveAllQueuedClassifications = (dispatch) => {
             console.info('Saved a queued classification, remaining:', queue.length);
             console.info('ducks/classifications.js submitClassification() success');
             dispatch({ type: SUBMIT_CLASSIFICATION_SUCCESS });
-            if (user) {
-              localStorage.removeItem(`${user.id}.classificationID`);
-            }
           } catch (error) {
             console.error('Failed to update classification queue:', error);
           }
@@ -182,10 +179,14 @@ const saveAllQueuedClassifications = (dispatch) => {
   dispatch(resetView());
 };
 
-const queueClassification = (classification) => {
+const queueClassification = (classification, user = null) => {
   const queue = JSON.parse(localStorage.getItem(FAILED_CLASSIFICATION_QUEUE_NAME)) || [];
   queue.push(classification);
+
   try {
+    if (user) {
+      localStorage.removeItem(`${user.id}.classificationID`);
+    }
     localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(queue));
     console.info('Queued classifications:', queue.length);
   } catch (error) {
@@ -259,7 +260,7 @@ const submitClassification = () => {
       },
       'metadata.subject_dimensions': subject_dimensions || [],
     });
-    queueClassification(classification);
+    queueClassification(classification, user);
     saveAllQueuedClassifications(dispatch);
   };
 };
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -10,7 +10,7 @@ import { resetView } from './subject-viewer';
 import { toggleDialog } from './dialog';
 import SaveSuccess from '../components/SaveSuccess';
 
-const CLASSIFICATIONS_QUEUE = 'classifications-queue';
+const CLASSIFICATIONS_QUEUE_NAME = 'classificationsQueue';
 
 //Action Types
 const SUBMIT_CLASSIFICATION = 'SUBMIT_CLASSIFICATION';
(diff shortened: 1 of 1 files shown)


snblickhan:
I'd label this a priority for beta 2 if _not_ having it means that the submit classification bug will still be happening. If there's another way around, that's fine -- whatever ensures we won't have volunteers losing work due to the length of the following subject.

wgranger:
This was not the cause of the bug (as we originally thought). The main question here is if we still need the queue.

snblickhan:
What does the queue give us, if not a bug fix?

wgranger:
If a classification fails, say a user is temporarily offline, or something unexpected happens, the failed classification will go into a queue, and that classification will attempt to resubmit again.

wgranger:
Ok, this should be all set, but I'd keep good eyes on it, because several different items were added to the classification duck since this was last rebased (already seen, save progress, Rollbar). I also moved the location of the `localStorage.removeItem(`${user.id}.classificationID`);` since putting it after the successful classification save in `saveAllQueuedClassifications()` caused the saved classification popup to arise when the next subject was loaded (even though the classification had already been submitted successfully).

src/ducks/classifications.js
```diff
@@ -121,10 +123,81 @@ const createClassification = () => {
...
+            dispatch({ type: SUBMIT_CLASSIFICATION_SUCCESS });
+          } catch (error) {
+            console.error('Failed to update classification queue:', error);
+          }
+          const { workflow, subjects } = actualClassification.links;
+          dispatch(addAlreadySeen(workflow, subjects));
+
+        })
+        .catch((error) => {
+          console.error('Failed to save a queued classification:', error);
```
shaunanoordin: We have two `console.error()`s here

src/ducks/classifications.js
```diff
@@ -121,10 +123,81 @@ const createClassification = () => {
...
+          console.error('ducks/classifications.js submitClassification() error: ', err);
```
shaunanoordin: 1. the `.catch()` argument is `error`, not `err` (should be named `err` for consistency)
2. This isn't `submitClassification()`, it's `saveAllQueuedClassifications()` (applies to other info logs in this function too)

src/ducks/classifications.js
```diff
@@ -121,10 +123,81 @@ const createClassification = () => {
...
+        .then((actualClassification) => {
+          console.log('Saved classification', actualClassification.id);
+
+          try {  //Fix: IE11 doesn't know what to do with Split.classificationCreated()
+            Split.classificationCreated(actualClassification);
+          } catch (err) { console.error('Split.classificationCreated() error: ', err); }
+
+          actualClassification.destroy();
+          const indexInQueue = queue.indexOf(classificationData);
+          queue.splice(indexInQueue, 1);
```
shaunanoordin: Here's the tricky thing; `queue.splice()` changes the very array that we're iterating through via `forEach()`.

Imagine...
```
var queue = ['a', 'b', 'c', 'd'];
queue.forEach(function (item, index) {
  console.log('item: ' + item + ', index: ' + index);
  queue.splice(index, 1);
});
```
...what do we expect the output to be?

A.
```
item: a, index: 0
item: b, index: 1
item: c, index: 2
item: d, index: 3
```

or B.
```
item: a, index: 0
item: c, index: 1
```

From what I can see, we _want_ the logic for `saveAllQueuedClassifcations` to iterate through _every_ item in the queue, so we want result A. However, `queue.splice()` is the equivalent of sawing the branch you're sitting on, so what we get is result B.

Basically, the in-place update of the `queue` has a risk of 'skipping' Classifications; what we need is a separate `queueToIterateThrough` and `queueToSaveBackToLocalhost`

shaunanoordin: e.g.
```
NEW_QUEUE = [];
queue.forEach((classificationData) => {
  apiClient.type('classifications').create(classificationData).save()
  .then((actualClassification) => {
    ...
    //If successful, don't do queue.splice()
    ...
  })
  .catch((err) => {
    ...
    //If UNSUCCESSFUL, return it to the queue
    NEW_QUEUE.push(actualClassification);  //Uh, gotta check if this function can actually see the actualClassification variable
  }
  ...
}

localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(NEW_QUEUE));
```

src/ducks/classifications.js
```diff
@@ -121,10 +123,81 @@ const createClassification = () => {
...
+        })
+        .catch((error) => {
+          console.error('Failed to save a queued classification:', error);
+          console.error('ducks/classifications.js submitClassification() error: ', err);
+          Rollbar && Rollbar.error &&
+          Rollbar.error('ducks/classifications.js submitClassification() error: ', err);
+          alert('ERROR: Could not submit Classification');  //TODO: better messaging
+          dispatch({ type: SUBMIT_CLASSIFICATION_ERROR });
+          switch (error.status) {
+            case 422: {
```
shaunanoordin: Is there any reason we're only triggering on a 422?

I mean, I see sense in this... maybe triggering on a 401 would cause an issue because maybe the user tried submitting when logged in, then logged out, then the re-submission tries to submit the Classification with invalid credentials.

But I just want to be sure we're catering for the pertinent possibilities. What if the user's WiFi drops? I'm not sure if 422 is the result we'll get.

eatyourgreens:
422 means the classification is invalid. If you look at the queue code for PFE, we drop invalid classifications but store all others to submit again later.

src/ducks/classifications.js
eatyourgreens: I think this code is correct. It deletes the classification if it was received by Panoptes but failed validation. 

A more elegant solution might validate the classification and repair it e.g.  missing required links or missing the `completed_at` timestamp.

src/ducks/classifications.js
eatyourgreens: This is a really good point. The code here seems to be copied from the PFE coffeescript, but I'm not sure if the PFE code is any good.

src/ducks/classifications.js
shaunanoordin: Thanks for the correction, I was reading the code wrong - I thought a Classification was only _re-queued_ when a 422 is received. Now I understand it's _removed_ from the queue when it's a 422.

src/ducks/classifications.js
```diff
@@ -121,10 +123,83 @@ const createClassification = () => {
...
+          console.log('Saved classification', actualClassification.id);
+
+          try {
+            Split.classificationCreated(actualClassification);
+          } catch (err) { console.error('Split.classificationCreated() error: ', err); }
+
+          actualClassification.destroy();
+
+          try {
+            localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, JSON.stringify(queue));
```
eatyourgreens: Why is the current queue being saved to local storage here?

eatyourgreens: I'm not sure if this should be `newQueue` or if you should delete this line completely.

shaunanoordin:
@wgranger we've got a timing/lifecycle issue.

Issue: The first Classification can be submitted fine, but second Classification can't, because it's `null`

Analysis:
- Prior to this PR, the life cycle for a Classification on submit would look like this, _in order:_
  - SUBMIT_CLASSIFICATION (in progress)
  - SUBMIT_CLASSIFICATION_SUCESS **(set `classification = null` to prepare for next Subject+Classification)**
  - FETCH_SUBJECT
  - FETCH_SUBJECT_SUCCESS
  - CREATE_CLASSIFICATION (`classification = {}`)
  - We now have a new Classification and Subject to work with.
- However, since we can now have _many_ SUBMIT_CLASSIFICATION_SUCCESS (...and FAILURES, but let's get to that later), it's tricky to know _when_ to trigger that FETCH_SUBJECT.
- In fact, this PR doesn't wait for SUBMIT_CLASSIFICATION_SUCCESS before calling FETCH_SUBJECT, so it becomes like this...
  - SUBMIT_CLASSIFICATION (in progress)
  - FETCH_SUBJECT
  - FETCH_SUBJECT_SUCCESS
  - CREATE_CLASSIFICATION (`classification = {}`)
  - Any combination of...
    - SUBMIT_CLASSIFICATION_SUCESS **(set `classification = null` to prepare for next Subject+Classification)**
    - SUBMIT_CLASSIFICATION_FAILURE
  - We now have a new Subject, but Classification `= null`

In fact, SUBMIT_CLASSIFICATION_SUCCESS and SUBMIT_CLASSIFICATION_FAILURE don't make sense any more with multiple classifications, since the latest status would overwrite the previous one.

src/ducks/classifications.js
shaunanoordin: Thanks for spotting this; this is incorrect. This line of code was originally used to... shrink the queue at each step, I think, and should be changed to something like...

```
const queue = JSON.parse(localStorage.getItem(FAILED_CLASSIFICATION_QUEUE_NAME));
localStorage.setItem(FAILED_CLASSIFICATION_QUEUE_NAME, /* null or '' or '[]' */ );  //Reset queue, then the queue will be refilled later on failed submissions.  
```

shaunanoordin:
## PR Update
This PR has been updated so the queue logic reflects this:
```
/*  saveAllQueuedClassifications() attempts to individually submit
    Classifications to Panoptes. What happens is that:
    - When a new Classification is meant to be submitted, it is placed in a
      queue via queueClassification()
    - Then, each item in the queue is individually processed.
      (The queue is emptied at this point, and a new, empty queue is prepared.)
      - If the item (Classification) is successfully submitted/saved, that's it.
        (It's effectively removed from the queue.)
      - If the item fails, it's reinserted into the new queue, to be processed
        the next time a new Classification is submitted.
    - Once ALL the items are processed, we are now ready to prepare our data
      for the next Subject to be fetched.
 */
```

ALSO:
- Each user gets his own queue. This allows users to reload the page in case of failure, and failed Classifications remain in the queue to be retried.

Tests OK on OSX+Chrome for:
- Submitting a Classification as a non-logged-in user
- Submitting a _second_ Classification right after the first.
- The above, but as a logged-in user
- Submitting a Classification after using the Save Progress feature as a logged-in user.

### Status
This PR needs to be re-reviewed. @wgranger please give this a look, and if it's OK, please merge & deploy.

src/ducks/classifications.js
```diff
@@ -51,20 +54,16 @@ const classificationReducer = (state = initialState, action) => {
       return Object.assign({}, state,{
         status: CLASSIFICATION_STATUS.SENDING,
       });
-
+      
     //Submitting Classification also resets the store.
-    case SUBMIT_CLASSIFICATION_SUCCESS:
+    //This is only called once the WHOLE queue of Classifications has been processed.
+    case SUBMIT_CLASSIFICATION_FINISHED:
```
wgranger: I can't find where this is called.

shaunanoordin: Hot damn, you're right, I didn't put that in.

src/ducks/classifications.js
```diff
@@ -186,38 +303,9 @@ const submitClassification = () => {
...
-        alert('ERROR: Could not submit Classification');  //TODO: better messaging
-        dispatch({ type: SUBMIT_CLASSIFICATION_ERROR });
-      });
-    //----------------
-
-    const { workflow, subjects } = classification.links;
-    dispatch(addAlreadySeen(workflow, subjects));
+    });
+    queueClassification(classification, user);
+    saveAllQueuedClassifications(dispatch, getState);
```
wgranger: MINOR: In `queueClassification(classification, user)` you pass in the user object. In `savedAllQueuedClassifications(dispatch, getState)` you pass in `getState` to immediately call `getState().login.user` in that function. Maybe go with `user` for consistency?

shaunanoordin: That makes a lot of sense. I'm on it.

src/ducks/classifications.js
```diff
@@ -121,10 +120,128 @@ const createClassification = () => {
...
+        //Right, why did it fail?
+        switch (err.status) {
+          case 422:  //If the reason for failure is that Panoptes returned a 422, it means the Classification was bad and should be discarded.
+            break;
+
+          default:  //Otherwise, the failed Classification should be re-queued for the next time attempt.
+            newQueue.push(classificationData);
+        }
+      })
+      .finally(() => {
```
wgranger: This Promise function has no IE support.

shaunanoordin: Whaaat? Isn't this polyfilled? Goddammit, let me check on my IE11, then if it effs up... I have a Plan B.

src/ducks/classifications.js
```diff
@@ -121,10 +118,128 @@ const createClassification = () => {
...
+          Split.classificationCreated(classificationObject);
+        } catch (err) { console.error('Split.classificationCreated() error: ', err); }
+
+        //Record locally that the Classification has been seen.
+        const { workflow, subjects } = classificationObject.links;
+        dispatch(addAlreadySeen(workflow, subjects));
+        
+        //OK, we don't need you any more, Classification. Why is this here? It's in PFE.
+        //TODO: check
+        classificationObject.destroy();
```
wgranger: I'm thinking this is here because we create another classification object with the same data, so we then have to delete the resource that was initially kept in the queue. However, I'm asking around in Slack.

src/ducks/classifications.js
wgranger: Doesn't look like it's currently polyfilled (unless it's somehow in the `es6` polyfill? I didn't test the failure, I'm just assuming based on what's on MDN. If there is a failure, it can easily be polyfilled https://polyfill.io/v2/docs/features/

src/ducks/classifications.js
```diff
@@ -121,10 +118,128 @@ const createClassification = () => {
...
+        }
+      })
+      .finally(() => {
+        //That's one item down.
+        itemsProcessed++;
+
+        //Have all items been processed?
+        if (itemsProcessed === itemsToProcess) {
+          //Did anything fail?
+          if (itemsFailed > 0) {
```
wgranger: What if two queued items fail? Will they get this alert twice?

wgranger: Oh wait, disregard this. I just noticed the indent when looking at this on Atom.

src/ducks/classifications.js
shaunanoordin: I ran a test using...

```
      .finally(() => {
        //That's one item down.
        itemsProcessed++;

        alert(itemsProcessed + ' of ' + itemsToProcess);
```

...on Win7+IE11 and it looks like it works. Let me know if you still see an issue on your end; if so, we can go with either

A. explicitly putting in that Polyfill for Promise.finally, or

B.
```
const finally = () => {
  ...
};

apiClient.type('classifications').create(classificationData).save()

//SUCCESS
.then(()=> {
  ...
  finally();
})

//FAILURE
.catch((err) => {
  ...
  finally();
});
```

src/ducks/classifications.js
shaunanoordin: 👌

wgranger: I can confirm this isn't deleting the desired classification though. I've confirmed that by submitted a classification on staging and later querying the API for the submitted classification ID, which returns the recently submitted item.

src/ducks/classifications.js
shaunanoordin: No, because the alert code will only be called when _all items_ have finished processing.

```
//Check that we're done with everything first
if (itemsProcessed === itemsToProcess) {
  //Did anything fail?
  if (itemsFailed > 0) { ...
```

src/ducks/classifications.js
wgranger: I just ran this on my end and all looked well.

shaunanoordin:
Update based on feedback:
- `SUBMIT_CLASSIFICATION_FINISHED` added
- `getState` replaced by `user` as a streamlining process
- `classificationObject.destroy();` question has been answered. (Will, how would you rewrite that comment for future clarification?)

Still waiting:
- confirmation for IE11 working with Promise.finally()

wgranger:
In this PR, `classificationObject.destroy()` does not delete the submitted classification from the database, it only destroys the copy. I can confirm the submitted classification still exists by querying the API for the classification ID after submission.

wgranger:
I've confirmed Promise.finally() is working on my end. If you have no other objections, I'm going to merge this and redeploy.

PR #179, opened 2017-11-22
Only Open Dialog without Selected Annotation

### LINKED ISSUE #167: When an Annotation is Selected, you can still select another Annotation
## Interaction Bug
- When Annotation X is selected, I can _still click_ on another Annotation, e.g. Annotation Y.
- This causes Annotation Y to be selected.
- Curiously, this causes two weird issues:
  - Although Annotation Y is selected, Annotation X's text will be shown.
  - If Annotation X has no text (empty string) and you switch to Annotation Y, Annotation X will be _saved as an Annotation with no text._

### Recommended Solution
- New logic needed: prevent `SelectAnnotation(objectThatIsNotNull)` if `store.selectedAnnotation !== null`

### Status
Medium-high priority. However, this shouldn't interrupt the Beta launch plans.

wgranger:
Fixes #167 

This fix was made via the suggestion on the issue.

### DIFF
--- a/src/ducks/annotations.js
+++ b/src/ducks/annotations.js
@@ -160,6 +160,11 @@ const selectAnnotation = (index, previousAnnotation) => {
     } else {
       annotation = getState().annotations.annotations[index];
     }
+
+    // If an Annotation is already selected, don't allow a new Annotation to be selected over it.
+    // Instead, use unselectAnnotation() to get a clean slate first.
+    if (annotation && getState().annotations.selectedAnnotation) return;
+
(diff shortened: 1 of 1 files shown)


shaunanoordin:
## PR Review
This PR prevents Annotations being selected (i.e. clicked on) when another Annotation is already selected (i.e. the Transcription dialog is showing), therefore preventing the nebulous "but wait who is truly selected?" behaviour.

### Status
👍  Noice! Approved, merging.

PR #181, opened 2017-11-23
Save Classifications for Later - Mk2

shaunanoordin:
## PR Overview
This PR is the second attempt at adding the feature which allows logged-in users to _manually_ save all the Annotations they've done, and resume Classifying at a later date.

This feature was originally added in PR #164 but reverted in PR #180 . The reason:
- When a user attempts to Submit a Classification, they'd receive _Error: Could not submit Classification_ with the following console error:
  - `PUT https://panoptes-staging.zooniverse.org/api/classifications/89036 422 (Unprocessable Entity)`
  - `Submit classification: Error -  Error: {"schema"=>"did not contain a required property of 'annotations'"}`
  - Request payload is:
```
{  
  "http_cache":true,
  "classifications":{  
    "completed":true,
    "metadata":{  
      "session":"6365883a0a1763274133432875d9320dcdf51fe0fac064e1c88efa50b4b058e9",
      "started_at":"2017-11-23T13:13:20.249Z",
      "user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36",
      "utc_offset":"0",
      "finished_at":"2017-11-23T14:57:22.642Z",
      "live_project":false,
      "user_language":"en",
      "user_group_ids":[  
        1326219
      ],
      "subject_dimensions":[  
        {  
          "naturalWidth":1962,
          "naturalHeight":1038,
          "clientWidth":1962,
          "clientHeight":1038
        }
      ],
      "workflow_version":"10.15",
      "viewport":{  
        "width":1363,
        "height":811
      }
    }
  }
}
```

This "second attempt" is intended to figure out wtf went wrong and fix the issue.

### Status
WIP

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "graphql-request": "^1.3.4",
     "history": "^4.6.1",
     "markdownz": "~7.3.1",
-    "panoptes-client": "^2.7.2",
+    "panoptes-client": "^2.8.0",
     "prop-types": "^15.5.10",
     "react": "^15.4.1",
     "react-dom": "^15.4.1",
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -73,7 +73,7 @@ class App extends React.Component {
     const showTitle = path === '/classify';
 
     return (
-      <div>
+      <div style={{ "position": "relative" }}>
         <Header />
         <ProjectHeader showTitle={showTitle} />
         {this.props.children}
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -73,7 +73,7 @@ class App extends React.Component {
     const showTitle = path === '/classify';
 
     return (
-      <div style={{ "position": "relative" }}>
+      <div>
         <Header />
         <ProjectHeader showTitle={showTitle} />
         {this.props.children}
--- a/src/components/ClassificationPrompt.jsx
+++ b/src/components/ClassificationPrompt.jsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import apiClient from 'panoptes-client/lib/api-client';
 import { connect } from 'react-redux';
 import { retrieveClassification } from '../ducks/classifications';
+import { fetchSubject } from '../ducks/subject';
 
 class ClassificationPrompt extends React.Component {
   constructor(props) {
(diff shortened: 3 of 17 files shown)


shaunanoordin:
Wait, here's a thought: what if, after a Classification object is created and `.save()d`, it can't be `.update()d` any more? What if you need to create a new Classification every time?

shaunanoordin:
Yeah, this hypothesis seems to have something to it.

```
const saveClassification = () => {
  return (dispatch, getState) => {
    ...    
    classification.update({
      annotations: [annotations],
      completed: false,
      'metadata.session': getSessionID(),
      'metadata.finished_at': (new Date()).toISOString(),
    })
    .save()  //HERE. As soon as this is called, any further attempts to submitClassification() will fail.
    ...
  };
};
```

shaunanoordin:
Eh, no dice. OK, here's what was attempted:
- Panoptes JavaScript client updated to 2.8.0. Result: 😢 
- After we Save Classifications In Progress, the `savedClassifiation` that's returned by Panoptes is then used to 'refresh' our current Classification object. Result: 😢 
  - (Also, SaveClassifications have been renamed to SaveClassificationsInProgress to avoid confusion)

OK, I need to shelve any further debugging on this until next week.

wgranger:
I was able to submit a classification by making a change to the `submitClassification` function in the classification duck. I added `annotations: classification.annotations` to https://github.com/zooniverse/anti-slavery-manuscripts/pull/181/files#diff-5b18d4fce53ac7c70f14c44cd56fed28R170, which seemed to fix the issue we were having. This is a similar fix that was made to your hypothesis above in the `saveClassificationInProgress` function. 

I'm not sure why such explicit changes to the annotation tag is needed with saved classifications. However, I think it has something to do with your idea that something changes when `.save()` is called and the classification is saved in our db. 

I can push up this simple revision, and you can take a look that the code is performing sound.

shaunanoordin:
Thanks to Will for pulling the debug! The problem was this: _you cannot change the Classification object on its own; you need to use the .update() function almost any time you're updating, say, classification.annotations._

I've updated code to make sure all (well, most - see the TODO) instances of, e.g. `classification.annotations.push(...)` is replaced with:
```
var updatedAnnotations = classification.annotations.slice();
updatedAnnotations.push(...);
classification.update({ annotations: updatedAnnotations });
```

👍  to Will for that bug spottin'.

Testing resumes now!

shaunanoordin:
Test updates:
- Non-logged-in user makes Annotation and submits: 👌 
- Logged-in user makes Annotation and submits: 👌 
- Logged-in user makes Annotations, Saves Progress, and submits: 👎 
  - We're getting a weird result here.

Say, a logged-in user...
1. Makes the Annotation "Hello"
2. Saves Progress
3. Makes the Annotation "Foo bar"
4. Submits

For some reason, the submitted Classification has _two_ instances of of the T1 task.

```
Payload to PUT https://panoptes-staging.zooniverse.org/api/classifications/89089 -

{  
  "http_cache":true,
  "classifications":{  
    "annotations":[  
      {  
        "task":"T1",
        "value":[  
          {  
            "frame":0,
            "points":[...],
            "details":[  
              {  
                "value":"Hello"
              }
            ]
          }
        ]
      },
      {  
        "task":"T1",
        "value":[  
          {  
            "details":[  
              {  
                "value":"Hello"
              }
            ],
            "points":[...],
            "frame":0
          },
          {  
            "details":[  
              {  
                "value":"Foo bar"
              }
            ],
            "points":[...],
            "frame":0
          }
        ]
      },
      {  
        "task":"T2",
        "value":1
      }
    ],
    "completed":true,
    "metadata":{  
      ...
    }
  }
}
```

Further tests needed:
- Logged-in user makes Annotation, Saves Progress. Close tab. Reopen tab, resume Progress. Make more Annotations, submit.

Wait, I know what's happening here:

shaunanoordin:
`const updatedAnnotations = (classification && classification.annotations) ? classification.annotations.slice() : [];  //Create a copy of the Annotations array so we can later update the Classification object with it.`

This was just added, to reflect all the times we used `classification.annotations.push()`

Turns out, what we should have done in every case is probably to start with NO annotations and build from scratch, because in all previous incarnations of ASM, we had assumed that `classification.annotations` would be empty `[]`.

OK, easy enough fix:

`const updatedAnnotations = [];  //Always start empty (don't pull anything from classification.annotation) the build the array based on the answers we have.`

shaunanoordin:
👍 

## PR Update
This PR has had several bugfixes in place to make sure SubmitClassification submits exactly what the user is seeing on their page. Of note:
- We avoid directly changing the Panoptes Classification object's properties where possible. (Dont' ask about `classification._workflow` or `classification._subject`; that's how it just works in [PFE.](https://github.com/zooniverse/Panoptes-Front-End/blob/f77ddbd227ab52069e472b4d542118998641f1de/app/pages/project/classify.cjsx#L162) )
- When submitting the Classification object, we build the `.annotations` array from scratch instead of using anything that already exists in `classification.annotations`. Because if we didn't, old 'saved' transcriptions would exist alongside all our new transcriptions.
- Submit Classification work OK in all tested scenarios.
- Save Progress continues to work.
- Continue Progress continues to work.

Tested 👌  OK for:
- Non-logged-in user makes Annotation and submits.
- Logged-in user makes Annotation and submits.
- Logged-in user makes Annotations, Saves Progress, and submits:
- Logged-in user makes Annotation, Saves Progress. Closes tab. Reopens tab, resumes Progress. Makes more Annotations, submits.

All the classification data submitted has been verified by checking the network data.

Debugging protip: for the first two test scenarios, data is PUT to `panoptes.zooniverse.org/api/classifications`, while the latter two scenarios, data is PUT to `panoptes.zooniverse.org/api/classifications/SUBJECT_ID`

### Status
@wgranger This PR is now ready for review.

wgranger:
Ok, I feel good about this PR after having both worked on it a bit. I think it's good we found that bug, as I feel more confident about how classifications are build since the most recent commits to this PR. I'm wondering if some of this work will also help with the mysterious classification submission errors we have been experiencing lately. 

Merging.

PR #183, opened 2017-11-24
Adjust Subject Query

wgranger:
The subject query was updated to achieve two changes.

1) Fix the initial bug of a subject query only pulling from the default workflowID (instead of the individual or collaborative workflow based on split group).

2) Allow a workflow to pull subjects from multiple subject sets. The current use case is that we are introducing a second beta subject set to allow the individual variant to view more subjects. This is achieved by randomly assigning a workflow one of its linked subject sets. We'll have to make sure the correct subject sets are linked through the project builder once this PR is merged.

*DO NOT MERGE until the second beta launch is ready


shaunanoordin:
I've taken a look at the code changes (no tests done yet) and I have two things on my mind:

1. Would the re-introduction of the "randomly select a Subject Set from the ones available" affect our ability to choose our Subject Sets in the Beta? Prooobably not, but I need to remember to test this since the //BETA_ONLY bits have been shuffled around.

2. Using `const workflow_id = getState().workflow.id || id;` is a functional way to set Workflow IDs for our project, but it makes me realise how awkward the `fetchSubject()` function is.

OK, let's break it down:

- `fetchSubject()` requires two parameters: `id` and `initialFetch`
- However, the line `const workflow_id = getState().workflow.id || id;` basically says, _"Yeah, if we know what our Workflow ID is, just ignore whatever id parameter is given to fetchSubject()"_
- Which in effect means the `id` parameter of `fetchSubject()` is _always_ ignored.
- _Which doesn't matter,_ because apparently `fetchSubject()` _never_ defines the `id` parameter anyway!
- I mean, search the code - `fetchSubject()` is only used twice - once on `SubjectViewer.componentDidMount()` for the initial fetch, and once in `classification.submitClassification()` for fetching subsequent Subjects.

With that in mind, my recommendation is to change `fetchSubject()` to remove the `id` parameter altogether, i.e. to

```
const fetchSubject = (initialFetch = false) => {
  ...
  const workflow_id = getState().workflow.id;
```

(Note: SubjectViewer.componentDidMount() would also need to be updated.)

src/ducks/subject.js
```diff
@@ -166,51 +166,40 @@ const selectSubjectSet = (id) => {
...
 
-    //Removed for //BETA_ONLY
     //----------------
-    /*
     let randomSubjectSet;
     const workflow = getState().workflow.data;
     if (workflow && workflow.links.subject_sets.length) {
       const linkedSets = workflow.links.subject_sets;
-      randomSubjectSet = linkedSets[Math.floor(Math.random()*linkedSets.length)];
+      randomSubjectSet = linkedSets[Math.floor(Math.random() * linkedSets.length)];
```
wgranger: This is where the second subject set is introduced. When constructing a query, a random linked subject set is chosen. In this use case, we would have both the first and second beta subject sets linked to the workflow, and one would be chosen at random.

shaunanoordin: Noted; thanks for pointing this out.

shaunanoordin:
## PR Review (Update)
Looking good! 👍 After the resolution of the merge conflict, this branch now functions properly. Website loads and core website functions seem to test OK.

### Status
Approved, merging.

PR #185, opened 2017-11-27
Hover Message on Gray Line

### LINKED ISSUE #176: Add Hover State to Grayed Out Lines
Add "this line has been transcribed!" hover state to grayed out lines to notify user to move along.

wgranger:
This branch is working, but there are a couple questions lingering during the review:

Currently, a user will not have a mouse down event over a previous annotation when in navigating or annotating mode. If this appears to throw off the user, I can pass those functions into a child prop to replicate their functionality. Just something to take note of during review.

Also made some line changes to appease the linter.

Closes #176

### DIFF
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -11,9 +11,9 @@ AnnotationsPane.jsx for details.
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Utility } from '../lib/Utility';
 import { connect } from 'react-redux';
 import { VisibilitySplit } from 'seven-ten';
+import { Utility } from '../lib/Utility';
 import PendingAnnotation from './PendingAnnotation';
 import { VARIANT_TYPES } from '../ducks/splits';
 import { MARKS_STATE } from '../ducks/subject-viewer';
@@ -32,7 +32,7 @@ class AnnotationsPane extends React.Component {
 
   render() {
     const pendingLine = this.props.mouseInViewer && this.props.annotationInProgress !== null;
-    const imageOffset = `translate(${-this.props.imageSize.width/2}, ${-this.props.imageSize.height/2})`;
+    const imageOffset = `translate(${-this.props.imageSize.width / 2}, ${-this.props.imageSize.height / 2})`;
     return (
       <g transform={imageOffset}>
         {pendingLine && (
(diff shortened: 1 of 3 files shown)


shaunanoordin:
👍  Overall, this PR looks great.

Regarding the "Retired Annotation Lines not responding to mouse events":
1. I agree that this behaviour would throw off the user's expectations, especially since the icon when you hover over a Retired Annotation Line is the Move/Pan/FourDirectionArrow icon.
2. Fortunately, the solution is pretty simple - here's a rough working version.
```
const renderedMarks = (
  <g
    className="annotation"
    key={annotationPrefix + index}
    style={style}
    onClick={(e) => {
      if (onSelectAnnotation) {
        onSelectAnnotation(index, previousAnnotations);
      }

      if (consensusLine) return;  //If retired line, don't stop events.
      return Utility.stopEvent(e);
    }}
    onMouseOver={(e) => {
      if (consensusLine) {
        this.tooltip.style.visibility = 'visible';
        return;  //If retired line, don't stop events.
      }
      return Utility.stopEvent(e);
    }}
    onMouseOut={(e) => {
      if (consensusLine) {
        this.tooltip.style.visibility = 'hidden';
        return;  //If retired line, don't stop events.
      }
      return Utility.stopEvent(e);
    }}
    onMouseMove={(e) => {
      if (consensusLine) {
        const cursor = this.props.getPointerXY(e);
        let rotationOffset;
        switch (this.props.rotation) {
          case 90:
            rotationOffset = 270;
            break;
          case 270:
            rotationOffset = 90;
            break;
          default:
            rotationOffset = this.props.rotation;
        }
        this.tooltip.setAttribute('transform', `translate(${cursor.x}, ${cursor.y}) rotate(${rotationOffset})`);
        return;  //If retired line, don't stop events.
      }
      return Utility.stopEvent(e);
    }}
    onMouseDown={(e) => {  //Prevent triggering actions in the parent SubjectViewer.
      if (consensusLine) return;  //If retired line, don't stop events.
      return Utility.stopEvent(e);
    }}
    onMouseUp={(e) => {
      if (consensusLine) return;  //If retired line, don't stop events.
      return Utility.stopEvent(e);
    }}
  >
    {svgLines}
    {svgPoints}
  </g>
);
```

Heck, that can even be simplified to `if (!consensus) return Utility.stopEvent(e);`

Further thoughts: the tooltip solution looks great (and an extra willgranger++ for compensating for transformations; I didn't even consider that) and it might be worth generalising - e.g. a tooltip for "click this last dot of this line to end your annotation!"

PR #187, opened 2017-11-27
Add Header to Dialogs

wgranger:
Not all headers are given a dialog (such as simple popups). However, the main use case here was the give the metadata popup a distinct header that stood out and stayed at the top when scrolling.

Additional line changes were made to keep the linter uniform.

classifier

### DIFF
--- a/src/components/Dialog.jsx
+++ b/src/components/Dialog.jsx
@@ -18,6 +18,19 @@ class Dialog extends React.Component {
     // }
   }
 
+  stopEvent(e) {
+    e.preventDefault && e.preventDefault();
+    e.stopPropagation && e.stopPropagation();
+    e.returnValue = false;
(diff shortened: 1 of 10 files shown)


src/components/Dialog.jsx
```diff
@@ -18,6 +18,19 @@ class Dialog extends React.Component {
     // }
   }
 
+  stopEvent(e) {
```
shaunanoordin: Couldn't we just use `Utility.stopEvent(e)` ?

shaunanoordin: Sorry, I only _just_ realised the code repetition when I saw `stopEvent()` being repositioned within this compo. Anyway, this should be marked as a minor comment - potential improvements on a later code cleanup.

PR #190, opened 2017-11-28
Retired Annotation Lines don't eat mouse events, plus stopEvent improvements

shaunanoordin:
## PR Overview
This PR makes some minor improvements to how we code Events.
- Retired Annotation Lines no longer eat mouse events, meaning we can now Pan (and Annotate)
 actions work when the mouse is over these greyed-out lines. (see PR #185 )
- Dialog and Popup both had their own versions of stopEvent(), which is actually a shared function available by the lib/Utility object. This PR standardises the use of stopEvent() in both components.

### Status
Ready for review.


wgranger:
Nice. Thanks for taking care of the consensus line issue and cleaning up the code a bit. LGTM.

PR #194, opened 2017-11-30
Functioning SVG Subject Loading Spinner

### LINKED ISSUE #137: Design: updating subject loading screen
Tangentially related to https://github.com/zooniverse/liberating-the-liberator/issues/121 

Instead of the blue circle inside the subject area, can we use this https://codepen.io/beckyrother/pen/BmjKWj centered in the grey box.

wgranger:
This PR places an SVG loading spinner in place of the previous pulsating circle. 

As always, a couple additional lines were modified per the linter's suggestions.

Closes #137 

screen shot 2017-11-30 at 11 38 24 am

### DIFF
--- a/src/components/SubjectLoading.jsx
+++ b/src/components/SubjectLoading.jsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SubjectLoading = ({ loaded }) => {
+  const transform = !loaded ? 'scale(0.35)' : '';
+
+  return (
+    <g id="Already-Seen-/-messaging" transform={transform} stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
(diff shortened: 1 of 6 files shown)


src/components/SubjectLoading.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SubjectLoading = ({ loaded }) => {
+  const transform = !loaded ? 'scale(0.35)' : '';
```
wgranger: This prop was added because there is an offset with how the parent's `<g>` transform changes once a subject is loaded. Look at the `SVGImage`'s parent `<g>` in the `SubjectViewer` and you'll notice there is a transform/scale that changes when `RESET_VIEW` is called in the `SVGImage`'s `onLoad` function. This shouldn't really be an issue since a loaded image will make the spinner disappear, but it was mainly added just in case.

PR #195, opened 2017-11-30
Hide classifier from non-admins

wgranger:
This is a simple PR to hide the classifier to non-admin while we are in a beta pause. The banner message was also changed. There are many ways to implement what is done in this PR, this seemed the best method I thought of.

### DIFF
--- a/src/components/Banner.jsx
+++ b/src/components/Banner.jsx
@@ -5,9 +5,8 @@ const Banner = () => {
   return (
     <div className="beta-top-banner">
       <h2>
-        Thanks for participating in our beta test.
-        Please click on the "Sign in" button at top right
-        to login or create an account.
+        We’re currently implementing some changes to the project based on
+        feedback from our volunteers. Stay tuned for our next round of beta testing!
       </h2>
     </div>
   );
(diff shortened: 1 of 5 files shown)


src/components/Banner.jsx
```diff
@@ -5,9 +5,8 @@ const Banner = () => {
   return (
     <div className="beta-top-banner">
       <h2>
-        Thanks for participating in our beta test.
-        Please click on the "Sign in" button at top right
-        to login or create an account.
+        We’re currently implementing some changes to the project based on
```
shaunanoordin: _Shakes fists angrily at slanted apostrophe_ Nyyyeeeaaarrrgh!

OK, this doesn't affect this PR in any way, but I need to complain about this. That curly apostrophe was an enemy of mine since the IE7/IE8 days because it's not a proper ASCII-safe `'` single quote used in coding, but a display-only character `’` formatted by word processors. It caused me hella problems back when browsers couldn't render it properly, but it's less of a problem in an era where _emojis_ are considered valid characters. 🤷‍♂️ 

In case you can't tell the difference between the two:

`It's Shaun's pet peeve`
vs
`It’s Shaun’s pet peeve`

Issue #188, opened 2017-11-28
Second Beta Prep.


wgranger:
TODO List for Second Beta:

[] Fix Subject queue bug with config error
[] Adjust feedback form to ask if user participated in initial beta
[] Deploy with cribsheet
[] Solve classification submission bug error

beckyrother:
**For the reminder:** the first time a user clicks Annotate, a modal comes up that's 390px wide (similar to the tutorial) with the gif from the tutorial showing where to place dots, and a paragraph:

> To create a new transcription, click below the start of the line of text just before the first word, then place one dot in each space between words.
> 
> When you reach the end of the line, place a dot under the space just after the final word.
> 
> Click the last dot again to open the Transcription box.

wgranger:
![screen shot 2017-11-30 at 3 38 10 pm](https://user-images.githubusercontent.com/14099077/33456263-7fb7345a-d5e4-11e7-8286-39ae95baf74d.png)
Like this?

snblickhan:
Can we add the "pause beta" option to this list, and prioritize it? Including a banner that reads "Thanks for participating in our beta review! We're in the process of implementing changes based on user feedback -- check back soon for our next round of testing!"

beckyrother:
Maybe a line at the end that says "check out the tutorial to learn more" and then never show it again?

snblickhan:
Can we make the URLs in the metadata clickable? This was a common point in the user feedback, that it was annoying to copy/paste those URLs instead of being able to just click on them.

![metadata url](https://user-images.githubusercontent.com/25958358/33573579-ca2f73ca-d8fb-11e7-9a92-0d779c619446.jpg)

snblickhan:
Also, I've created new feedback forms for Beta #2. 

Independent: https://goo.gl/forms/h0V82sJQ4zEmveCH3
Collaborative: https://goo.gl/forms/7Ur0kCxD1ZIPjNgA2

snblickhan:
- [ ] “DIRECTIONS” on classifier needs to be updated. Should read “Using the Annotate tool, click under the start and end of a line of text, then add your transcription. [2nd para is fine as is].”

- [ ] Using M and A for toggle buttons isn’t going to work — whenever you type them as you’re transcribing they’re triggering the shortcuts, which is really confusing/annoying.

[...]

PR #196, opened 2017-12-04
Reminder on First Annotation

wgranger:
Per suggestion in #188, this PR adds a reminder to the user when making a first annotation. After the reminder is seen, a tag is saved in the user preferences.


snblickhan:
@beckyrother Do you think it's worth keeping this reminder if we're going to endpoint dot annotation? My vote would be yes, but I'm curious to know what you think.

beckyrother:
Yes, I think it's still useful, especially as we're changing the way this works.

shaunanoordin:
Quick check: do we still need this Reminder on the First Annotation popup? If yes, there's a bit of merging and/or rebasing that I should get on.

wgranger:
Ooo. Didn't notice that rebase needed until you point it out. I believe we still need this. I'll put a post in the channel.

shaunanoordin:
Quick thoughts:
- I've confirmed that the reminder appears properly for a non-logged-in user, but testing for a registered user has been an issue due to the saved preferences. All said, I'm happy enough with the implementation.
- Visual design looks good. 👍  Was going to write a report about it looking bad but then I remembered that when switching to a new branch, new stylesheets aren't incorporated until you restart the dev server. Derp.

snblickhan:
@shaunanoordin Apologies -- I updated this modal text over in the Second Beta Prep thread a few days ago, as Tom Blake wanted a tutorial update about words split across two lines. I tagged Will & Becky, but not you, so this is 100% on me. Any way we can add the additional text (shown in bold below)?

> To create a new transcription, click below the start of a line of text just before the first word, and again just after the last word. This will underline the text you wish to transcribe. Please annotate and transcribe entire lines at a time. **Words split between two lines should be typed out in their entirety on the first line.** If you're unsure of anything, please revisit the project tutorial.

wgranger:
Sure. This can be done easily. @snblickhan

PR #198, opened 2017-12-15
Convert to "Single Line" Annotations

shaunanoordin:
## PR Overview
This is a major PR that changes the way we create Annotations. We're moving from multi-point annotations (i.e. one 'dot' for each word and the end of the line of text) to "single line" annotations (i.e. one dot at the start of the line of text, one at the end)

(Becky, if you're reading this because I tagged you, the tl;dr for you is that you need to give this the Design OK after we've done the code changes. Everything else is just code jibba jabba.)

### The Plan
Generally, three things:
1. _Temporarily_ disable all Previous Annotations until the Caesar aggregation is updated.
  - `ducks/previousAnnotations` will just be hacked so that it assumes no results whatsoever.
2. Annotations should end after only two dots.
  - Also, in SelectedAnnotation, disable word count logic.
  - OPTIONAL (recommended?): revise data structure of an Annotation.
3. _Re-enable_ Previous Annotations WHEN Caesar aggregation is updated.

### Handy Checklist
- [x] When creating a Annotation, the second click/dot should trigger `completeAnnotation()`. (Done in SubjectViewer.jsx)
  - [x] Annotations In Progress should ignore angle limits.
  - [x] Annotations In Progress should no longer recognise clicking on the "last dot" to complete the Annotation.
- [x] The SelectedAnnotation component should, like, totally chill out with the word count logic. User can submit any String data... as long as it's not empty.
  - [x] Be aware of the special case where a user Cancelling out of a newly created Annotation should also Delete that new Annotation. (See below for discussion on how the Annotation data is stored.)
  - [x] NEW: When clicking DONE on a Selected Annotation with blank text, that Annotation is deleted. Remember, previously, it was impossible to submit Annotations with blank text due to the word count system, so we need to compensate for the removal of said word count. 
- [x] User should be able to submit a Classification object that makes sense to the researchers
  - [x] ...and to Coleman's aggregator
```
annotations.annotations: [
  {
    details: ["Hello", "World"],
    frame: 0,
    points: [{x:0, y:0}, {x:0, y:100}]
  }, ...
]
```
- Looking above at how the Annotation data is stored...
  - ...we can see that `.frame` is fine. `.points` can remain as-is, since we can sorta safely assume that all our annotations are lines with exactly two items (x-y coords) in the 'points' array. (Besides, changing this to `.startPoint` and `.endPoint` will take more work than necessary.)
  - ~~...however, `.details` no longer makes sense being stored as an array. 💀~~
    - ~~Recommendation: Check with Coleman how he'd like to have the data for the aggregator.~~
    - ~~By default though, I'll _replace_ `.details` with a new field called `.text`, which stores a single String.~~
  - (UPDATE) ...however, `.details` will no longer be an array of words.
    - Instead, `.details` will be an array with a _single_ item, and that item will be a full line of transcribed text.
    - e.g.: `{ details: ["I like big bugs and I cannot lie, you other entomologists can't deny
"] }`
    - See the [discussion below](https://github.com/zooniverse/anti-slavery-manuscripts/pull/198#issuecomment-352452555) for details on the 'correct' data structure.
- [ ] Tutorials and text need to be updated. (Check with Sam?)
- [ ] Once done, check with @beckyrother for final pass on new UI implementation.

### Status
WIP. This PR will be worked on by either Shaun or @wgranger

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -22,7 +22,6 @@ class SelectedAnnotation extends React.Component {
     this.deleteAnnotation = this.deleteAnnotation.bind(this);
     this.insertTextModifier = this.insertTextModifier.bind(this);
     this.cancelAnnotation = this.cancelAnnotation.bind(this);
-    this.deleteAnnotation = this.deleteAnnotation.bind(this);
 
     this.state = {
       annotationText: '',
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -22,7 +22,6 @@ class SelectedAnnotation extends React.Component {
     this.deleteAnnotation = this.deleteAnnotation.bind(this);
     this.insertTextModifier = this.insertTextModifier.bind(this);
     this.cancelAnnotation = this.cancelAnnotation.bind(this);
-    this.renderWordCount = this.renderWordCount.bind(this);
 
     this.state = {
       annotationText: '',
@@ -172,8 +171,6 @@ class SelectedAnnotation extends React.Component {
       width: PANE_WIDTH,
     };
 
-    const wordCountMatchesDots = this.doesWordCountMatchDots(this.state.annotationText, this.props.selectedAnnotation.points.length);
-
     return (
       <Rnd
         default={defaultPosition}
@@ -218,12 +215,8 @@ class SelectedAnnotation extends React.Component {
             this.renderAnnotationOptions()
           )}
 
-          {!this.state.showAnnotationOptions && (
-            this.renderWordCount(this.state.annotationText)
-          )}
-
           <div className="selected-annotation__buttons">
-            <button className="done-button" disabled={!wordCountMatchesDots} onClick={this.saveText}>Done</button>
+            <button className="done-button" onClick={this.saveText}>Done</button>
             <button onClick={this.cancelAnnotation}>Cancel</button>
             {(this.props.annotation.previousAnnotation) ? null :
               <button onClick={this.deleteAnnotation}>Delete</button>
@@ -264,26 +257,6 @@ class SelectedAnnotation extends React.Component {
         })}
       </div>
     )
-  }
-
-  renderWordCount(text) {
-    const expectedWords = this.props.selectedAnnotation.points.length - 1;
-    const cleaned_text = text.replace(/\s+/g, ' ').trim();
-    const number_of_words = cleaned_text ? cleaned_text.split(' ').length : 0;
-    const style = expectedWords === number_of_words ? "word-count--green" :
-      expectedWords < number_of_words ? "word-count--red" : "";
-
-    return (
-      <span className={`word-count ${style}`}>
-        {number_of_words} &#47; {expectedWords} words
-      </span>
-    )
-  }
-
-  doesWordCountMatchDots(text, dots) {
-    const cleaned_text = text.replace(/\s+/g, ' ').trim();
-    const number_of_words = cleaned_text.split(' ').length;
-    return number_of_words === dots - 1;
   }
 
   saveText() {
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -22,6 +22,7 @@ class SelectedAnnotation extends React.Component {
     this.deleteAnnotation = this.deleteAnnotation.bind(this);
     this.insertTextModifier = this.insertTextModifier.bind(this);
     this.cancelAnnotation = this.cancelAnnotation.bind(this);
+    this.deleteAnnotation = this.deleteAnnotation.bind(this);
 
     this.state = {
       annotationText: '',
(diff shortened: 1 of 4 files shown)


shaunanoordin:
Alright @wgranger , the basic changes into single-line Annotations are in place. The Classification data submitted, incidentally, looks like this:

```
{  
  "http_cache":true,
  "classifications":{  
    "annotations":[  
      {  
        "task":"T1",
        "value":[  
          {  
            "details":[  
              {  
                "value":"There being no time for the description of the Revolution"
              }
            ],
            "points":[  
              {  
                "x":975.3602693602693,
                "y":447.986531986532
              },
              {  
                "x":1579.077441077441,
                "y":447.986531986532
              }
            ],
            "frame":0
          }
        ]
      },
      {  
        "task":"T2",
        "value":1
      }
    ],
    "metadata":{...}
}
```

There's always exactly two items in the `.points` Array - the start and end points. The format of the `.data` variable sorta... kinda... sorted itself out? I'm not sure how that happened, but it's always an Array with a single item, which is the full line of transcribed text. Yay! (Yay?)

CKrawczyk:
@shaunanoordin I was just about to post that you had the wrong data structure for the annotation in the opening post.  The structure you show now is the correct one and does not need to change for the new reducer to work.

shaunanoordin:
Wait, that structure works? Sweet! High five, @CKrawczyk ! 😄 ✋

CKrawczyk:
Yep, the extractor does not need to change at all, just one flag changing on the reducer.  I can change this now, although it might make things look odd coming out of the reducer since the Caesar database needs to be cleared for the workflows.

shaunanoordin:
That's cool - users are currently locked out of the ASM website as we manage these updates, so we're safe to have a transition period where the data/front-end/etc can be wonky. We can coordinate this transition over Slack. 👌

wgranger:
The structure above looks nice. Only note, you mentioned earlier changing `details` to `text`. However, I believe the structure above should be used instead of changing the `details` tag to `text` because the Panoptes backend expects this data structure when making a data export for the polygon tool.

shaunanoordin:
Ah, good point @wgranger - I'll update the comment in the first post accordingly.

shaunanoordin:
## PR Update
Thanks to @CKrawczyk's updates on Caesar (staging), we can now properly receive aggregated data. 👍 

- External dependency: Caesar (staging/production)
- Only new Classifications (made today onwards) count towards creating Caesar Reductions (i.e. aggregated results)
- The aggregated data from Caesar WORKS AUTOMAGICALLY. Apparently no changes are needed in our front end code to get this working, woot woot.
  - However, **NOTE:** we may need to double-check what happens with excess white space. i.e. aggregated data will likely have its white space trimmed, so what happens when a user submits `"    Gosh     that's a     lot   of  spaaaace      !     "` ?
- The aggregated data can be tested on (staging) Workflow 3085 Subject 51842 (See the [Caesar page](https://caesar-staging.zooniverse.org/workflows/3085/subjects/51842) ) which now has a few successful "Previously Annotated (by Other People) Lines" showing.

screen shot 2018-01-04 at 17 47 15
_The Previously Annotated Lines on Workflow 3085, Subject 51842_

screen shot 2018-01-04 at 17 47 04
_The text from previous users' annotations are showing properly, noice_

### Status
Ready for review, @wgranger .

Due to how important this code (functionality) update is, I'll separate the content-related subtasks to another PR. So we'll need to approach the following separately:
- [ ] Tutorials and text need to be updated. (Check with Sam?)
- [ ] Once done, check with @beckyrother for final pass on new UI implementation.

src/components/SelectedAnnotation.jsx
```diff
@@ -22,7 +22,7 @@ class SelectedAnnotation extends React.Component {
     this.deleteAnnotation = this.deleteAnnotation.bind(this);
     this.insertTextModifier = this.insertTextModifier.bind(this);
     this.cancelAnnotation = this.cancelAnnotation.bind(this);
-    this.renderWordCount = this.renderWordCount.bind(this);
+    this.deleteAnnotation = this.deleteAnnotation.bind(this);
```
wgranger: This is a repeated line from above.

wgranger:
Overall, this looks pretty good. However, it seems a bit odd that this information is returned from Caesar (which is immediately added to the "data" tag in the previousAnnotations redux store. 

screen shot 2018-01-04 at 2 36 51 pm

beckyrother:
Talked to Will IRL but this looks good on my end. Thanks!

CKrawczyk:
@wgranger The `data: {"": 1, 1: 5}` bit is the reducer for the "is this done" task (note that is back filling all annotations, not just starting from yesterday.  A min workflow number will need to be set for this extractor on caesar to make sure this does not happen on production).

The `data: {"frame0": []}` bit is the old reducer from the previous beta.  I needed to make a new reducer label in caesar to prevent the old data format from being combined with the new one.  Since the production database has be cleared out this will not be an issues on production.

CKrawczyk:
As a side note adding `reducerKey:"ext-17"` to line https://github.com/zooniverse/anti-slavery-manuscripts/blob/3d523526318304657d862ffaaefe9b7f43fd9692/src/ducks/previousAnnotations.js#L104 will limit the returned output to just the new reducer.  Although that will only be for staging, it will need to be `reducerKey:"ext"` for production.

PR #199, opened 2017-12-20
Keep Annotation Box Within Window Bounds

### LINKED ISSUE #79: Set Annotation Box Boundaries
Currently, the annotation box can display out of bounds, which is bad UI. It would be nice if the box remained contained to the window.

wgranger:
Fixes #79 

This PR creates a function that determines where the annotation box boundaries will lay and adjusts the box is out of bounds.

CSS styling was changed from margin to padding because `offsetWidth` and `offsetHeight` don't take margins into account.


snblickhan:
This would be nice to have for beta 2, since people gave us feedback about it in the first beta.

shaunanoordin:
Goddammit merge conflict! Let's see what's up here...

PR #212, opened 2018-01-09
IE11 fix: allow Classifications to be submitted

shaunanoordin:
## PR Overview
This PR fixes an issue with IE11, where - currently - if you attempt to submit a Classification (Finish -> pages fully done/not fully done -> Done), you'll receive an alert that says "ERROR: Could not submit Classification" DESPITE the fact that the Classification was submitted (as verified by console and network logs.)

The problem was that - for some mysterious reason - IE11 doesn't know how to handle `Split.classificationCreated(classification)` in `submitClassification()`, so a simply try-catch was placed around it.

### Status
Ready for review, @wgranger

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -186,7 +186,9 @@ const submitClassification = () => {
       }
       //Log
       console.log('Submit classification: Success');
-      Split.classificationCreated(classification);
+      try {  //Fix: IE11 doesn't know what to do with Split.classificationCreated()
+        Split.classificationCreated(classification);
+      } catch (err) { console.error('Split.classificationCreated() error: ', err); }
 
       //Reset values in preparation for the next Subject.
       dispatch({ type: SUBMIT_CLASSIFICATION_SUCCESS });


src/ducks/classifications.js
```diff
@@ -186,7 +186,9 @@ const submitClassification = () => {
       }
       //Log
       console.log('Submit classification: Success');
-      Split.classificationCreated(classification);
+      try {  //Fix: IE11 doesn't know what to do with Split.classificationCreated()
+        Split.classificationCreated(classification);
```
wgranger: If merged, I'm not sure this should be a permanent fix. It's important we have this `.classificationCreated` function to track classifications through seven-ten. I have a suspicion there could be a JS function somewhere in `classificationCreated()` that isn't supported in IE? However, I feel like this should appear somewhere in the log.

wgranger:
I think this is a good temporary fix, but it does arise a larger problem of why classifications aren't logged in seven-ten through IE. That said, this fix is better than the UI error issue popping up, which isn't logging the classification to seven-ten anyways.

I'm merging this, but I'm also opening an issue to try to resolve this before the full launch.

Issue #216, opened 2018-01-11
User cannot submit Classification


shaunanoordin:
## Issue needs investigating
A user encountered an error message at the submission process of a fully transcribed/annotated letter.

- Original report: https://www.zooniverse.org/projects/bostonpubliclibrary/anti-slavery-manuscripts/talk/1183/477165
  - **Letter finished but system would not accept it**
  - _Finished a letter but had a one hour gap before the last page. The last page had problems (post mark, different orientation of inscription etc) and when I finished it would not accept the letter. Are all the four pages of work wasted - I hope not!_
- Worth double-checking our previous fix. https://github.com/zooniverse/anti-slavery-manuscripts/pull/192
- Problem is, this is such a vague error that once again we're searching for a noodle in a haystack. (We could have used a magnet if it was a needle.)
- **We may need an error logging system.** Shaun: check Rollbar that Cam invited you to.
- The queue system ("send Classifications later") may be a way to pave over visible errors, but we may still need to find the root cause

shaunanoordin:
Update: Fix #192 seems to be in place, so this bug is an exciting new pain in the butt to debug.

wgranger:
Note: I've tried transcribing on the final page of a subject (with the next subject being less pages) to see if we had a similar error as that from #192, but everything looked alright there.

A potential step in the right direction was to try logging errors to GA.

PR #219, opened 2018-01-15
Utilize Rollbar to Detect Classification Submission Errors

wgranger:
This PR adds Rollbar functionality to the app, getting us one step closer to diagnosing and closing #216. 

My main questions are:
- `Rollbar.configure()` is used in the `App` component to determine the environment, as it was a bit tricky to determine the environment through `process` and conditionals on the html script. 
- I'm also wondering if there's a way to minify that huge code block in `index.tpl.html`.
- Currently, there's a good bit of logging going on with the classification catch, so maybe we should have a discussion of what we want going on here.


shaunanoordin:
First question: have we successfully logged any useful error messages?

I went to rollbar.com and saw that the recent (testing) error messages only shows... `TestRollbarError: testing window.onerror File http://localhost:3000/classify line (unknown) in [anonymous]`
...which isn't too helpful.

Thoughts:
1. do we need to submit our ASM sourcemap to Rollbar, and if so, how? (Sounds like a LMGT)
2. Do we need to expand any error objects? I saw the [documentation](https://rollbar.com/docs/notifier/rollbar.js/) for Rollbar.js but I wanted to confirm that `Rollbar.error("Goddammit", err)` catches the `err` object correctly (which, honestly, may just be a string), and we don't need to do something like `Rollbar.error("Goddammit", JSON.stringify(err_object))` to read out the full details of the error.

src/components/App.jsx
```diff
@@ -48,6 +48,7 @@ class App extends React.Component {
   componentDidMount() {
     this.props.dispatch(fetchProject());
     this.googleLogger.remember({ projectToken: 'antiSlaveryManuscripts' });
+    Rollbar.configure({ payload: { environment: env } });
```
shaunanoordin: I'm wondering if we should also configure to record which user is reporting the error; this might help us with user comments on Talk.

But... that might be tricky since the User login check happens at a much later stage in the app life cycle. Can you run `Rollbar.configure()` _again_ at a later stage to add/edit values?

src/index.tpl.html
```diff
@@ -6,6 +6,17 @@
     <title>Anti-Slavery Manuscripts</title>
     <script src="https://ft-polyfill-service.herokuapp.com/v2/polyfill.min.js?features=default,Array.prototype.includes,Array.prototype.keys"></script>
 
+    <script>
```
shaunanoordin: You know what? This seems cool. I wouldn't be worried about minifying it, because it's third party code and it's good to have eyes on what we added.

src/components/App.jsx
wgranger: Yes, you can add `Rollbar.configure()` at a later stage to adjust the params however you want.

wgranger:
- When you run something like `Rollbar.error("Bleh", err)`, the `err` will return in the email as `body.message.extra.extraArgs`; however, it is still work running a classification error to see exactly what that `err` looks like.

shaunanoordin:
@wgranger 

screen shot 2018-01-20 at 00 50 21

1. Turns out `Rollbar.error("ErrorTypeBleh", errData)` works quite well; if `errData` is an object, the full object data is recorded in Rollbar - you just have to dig around a bit to find it.
  - Go to Rollbar -> click on the 'Error Group/Error Type' (e.g. "ErrorTypeBleh") -> click on Occurrences tab -> each occurrence of Bleh is listed with all the errData's values. (It'll be a super wide table if errData is an object). Bonus: -> click on a timestamp to open a specific Occurrence item to see full telemetry and params.
2. Holy butts, I just realised that when you call `Rollbar.error()`, it grabs a telemetry of: A. most network calls, B. most (not all) user actions, C. a large chunk (of almost useless) log entries.

Log entries, as in all this nonsense in the browser dev console:
screen shot 2018-01-20 at 00 55 13
...which includes unnecessary data like font colour, but lacks any pertinent data because each console entry is _truncated._

During my tests (where I forced an error on Classification), I had a huge chunk of nonsense telemetry recorded, and my test ran only for 60 seconds. Right - how much data are we recording when a user has been on the page for 60 minutes?

wgranger:
Ohh. I didn't realize that. It seems rather odd that Rollbar would log all that info, so I'm thinking there has to be a way to adjust that. We definitely don't want a telemetry of everything logged from a 30+ minute session.

shaunanoordin:
Three things:
- [ ] I'll try resolving the merge conflict via Github web editor
- [ ] I'll see if I can disable the Redux console.log() via `editing webpack.production.config.js` - this should significantly reduce telemetry data. Plus, our public users wouldn't find this useful anyhoo.
- [ ] I'm going to Ctrl+F all instances of try-catches that don't output some sort of error message.

On that last point, let me know if you spot a problem with this plan:
- For non-function-breaking errors, `console.error()` should be enough to track the problem (which might then be relevant to a breaking error)
- For function-breaking errors, `console.error()` and `Rollbar.error` to get everything, including minor errors reported previously.

wgranger:
I've already got a PR up to cease Redux logging on production (#232), but feel free to make any suggestions or try a different approach if you find a better solution. 

I think `Rollbar.error()` sounds like a good approach, but we should test it out one final time to ensure we are getting the error info we are looking for with the log.

shaunanoordin:
Ah, I should have refreshed my PR page. Thanks @wgranger , I'll get that reviewed. 👍

shaunanoordin:
One thing I just noticed is that the `Rollbar` object isn't `import`-ed into the script... it's a global object? i.e. via the `<script>` tag in the HTML `<body>`?

If so, this requires some thought - is there any possibility that the `Rollbar` object won't initialise properly? e.g. if the Rollbar src file isn't available for whatever reason, and `window.Rollbar = null` or `window.Rollbar = {}`? i.e. we may not be able to assume that `Rollbar.error()` is a valid object.method() call at all times, and if Rollbar or Rollbar.error is ever null, the whole app may crash _trying to report an error._

Following that thought: is `Rollbar && Rollbar.error && Rollbar.error('...')` enough of a safety?

shaunanoordin:
@wgranger I've updated a whole lot of the error messaging, and I'd like to get your feedback on the implementation. I _think_ we're tracking everything that needs to be tracked, but I can't shake the fact that I may be missing something.

wgranger:
Ok. I've given this another look and all appears well. I like the safety addition of the Rollbar checks as well as the tag of _where_ the error is coming from in the ducks. LGTM.

PR #225, opened 2018-01-17
[RFC] Remove Beta Mentions

wgranger:
This branch removes all mentions of beta in both the code and UI. The PR is currently RFC because there are a few items to discuss.

- Although I understand why `login.initialised` is present in the ducks, I'm not sure it is changing much in the frontend logic. 
- Although the top banner was initially marked `BETA_ONLY`, I think it makes sense to keep it in if we want to make an announcement in the future.
- It's possible we will need subject set selection once more subject sets are introduced. For now, the comments were changed to `INITIAL_LAUNCH_ONLY`

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -82,7 +82,6 @@ class App extends React.Component {
       <div>
         <Header />
 
-        {/*//BETA_ONLY: Notify of beta and feedback form*/}
         {(this.props.showBanner)
           ? <Banner hideBanner={this.hideBanner} /> : null
         }
--- a/src/components/Banner.jsx
+++ b/src/components/Banner.jsx
@@ -1,10 +1,9 @@
-//BETA_ONLY
 import React from 'react';
 import PropTypes from 'prop-types';
 
 const Banner = ({ hideBanner }) => {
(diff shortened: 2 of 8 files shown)


shaunanoordin:
Some quick thoughts:

> Although I understand why login.initialised is present in the ducks, I'm not sure it is changing much in the frontend logic.

Generally, I consider the `initialised` value to signal that a component is ready to actually do its work, if it needs to check if a user is logged in. (Basically, we should rewrite this as `user_status === USER_STATUS.READY`

For example, we if we ran this code: `if (!this.props.user || !this.props.user.hasSeenTutorial) { renderTutorial() }` there will be a split second when a logged-in user will see the tutorial before it disappears, since the app starts with `this.props.user` being `null` before changing it to an object.

> Although the top banner was initially marked BETA_ONLY, I think it makes sense to keep it in if we want to make an announcement in the future.

👍  Agreed. `<Banner>`, you have been promoted.

> It's possible we will need subject set selection once more subject sets are introduced. For now, the comments were changed to INITIAL_LAUNCH_ONLY

👍  Sweet.

PR #226, opened 2018-01-19
Add Already Seen Banner

### LINKED ISSUE #110: Already Seen?
This is more a design question. The "Already Seen" label is standard in PFE, but we haven't considered yet if it should be added to ASM.

### LINKED ISSUE #218: Implement "Already Seen" banner
## Feature Required
When the ASM Classifier serves a Subject that the user has already seen, we need some sort of banner/sign/etc indicating this is the case.

### Status
Medium priority. Doesn't affect functionality, but it assuages users' concerns when (after the ASM runs out of new Subjects to serve) the users see a document they've already spent hours classifying prior, but with their work apparently "lost". (Explanation: their work isn't lost; it has already been submitted to Panoptes and therefore doesn't need to be shown ala Saved Progress Classifications.)

wgranger:
This PR adds the Already Seen banner atop a subject if the user has already submitted a classification for the subject or seen it during the session. 

Adding the banner pushes the subject viewer down a bit, so a hack was done in the FilmstripViewer that adds more to the height in such a case. 

The only difference between this banner and the one icon in PFE (aside from appearance) is the addition of the "What's This?" button which elicits a popup with a quick description. I'm open to suggestions on the content of that popup.

Closes #218 
Closes #110


src/components/FilmstripViewer.jsx
```diff
@@ -47,7 +47,7 @@ class FilmstripViewer extends React.Component {
     const images = getAllLocations(this.props.currentSubject);
 
     const render = images.map((image, i) => {
-      const activeBorder = i === this.props.frame ? "related-images__frame--active" : "";
+      const activeBorder = i === this.props.frame ? "related-images__frame--active" : '';
```
shaunanoordin: [super minor] Waaait a minute why is an empty string getting the single quote treatment but not the string `"relate-images__frame--active"`? Are there some linting shenanigans going on?

src/lib/seen-this-session.js
```diff
@@ -0,0 +1,26 @@
...
+    subjectsSeenThisSession.push(`${workflowId}/${subjectId}`);
+  });
+}
+
+function checkAlreadySeen(workflow, subject) {
+  return subjectsSeenThisSession.includes(`${workflow.id}/${subject.id}`);
+}
+
+if (window) {
+  window.subjectsSeenThisSession = subjectsSeenThisSession;
```
shaunanoordin: This works, but why are we saving the array of Subjects seen (in one session) into the global namespace instead of chucking said array into a duck? I'm not opposed to using global variables, but I'd like to know why we're breaking with our usual data store pattern.

src/lib/seen-this-session.js
```diff
@@ -0,0 +1,26 @@
+import auth from 'panoptes-client/lib/auth';
+
+const subjectsSeenThisSession = [];
+
+auth.listen('change', () => {
+  return subjectsSeenThisSession.splice(0);
+});
+
+function addAlreadySeen(workflowId, subjectIds) {
```
shaunanoordin: Wait, hang on, is this ever called?

src/components/FilmstripViewer.jsx
wgranger: Oh wait. Haha. Yeah, I didn't notice the strange different with that. The linter is kinda weird, so I'll take a look or change it back.

src/lib/seen-this-session.js
wgranger: No reason really, this was just how it was done in PFE. For our use, it probably would make more sense to put it in a duck, which I can get done in another commit.

src/lib/seen-this-session.js
wgranger: Oh, it should've been in the classifications duck. I might've forgotten to save that file before pushing the commit.

PR #231, opened 2018-01-22
Allow IE to Log Seven-ten

### LINKED ISSUE #214: Seven-ten Isn't Logging in IE
IE has a problem with the `Split.classificationCreated` function, and a temporary fix was found in #212. However, we need this function to log seven-ten classifications in the project. Before full launch, a fix should be found.

wgranger:
My initial testing of this branch, after the addition of `es6` to the polyfill, appears to solve the issue of logging `Split.classificationCreated()`, but my console access w/ my IE VM is a little spotty, so it's worth a further review before any merge.

Fixes #214

### DIFF
--- a/src/ducks/previousAnnotations.js
+++ b/src/ducks/previousAnnotations.js
@@ -123,7 +123,6 @@ const fetchPreviousAnnotations = (subject) => {
       });
     })
     .catch((err) => {
-      console.log(err);
       dispatch({ type: FETCH_ANNOTATIONS_ERROR });
     });
   };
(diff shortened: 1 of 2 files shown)


shaunanoordin:
OK, I'll test this on IE11 and- wait whaaa? The polyfill was just moved to the bottom of the HTML?

wgranger:
No, `es6` was also added to the polyfill.

shaunanoordin:
Ah, I need to read that line better. Thanks, Will!

PR #238, opened 2018-01-24
More Error/Status logging and remove shift() from retrieveClassification()

shaunanoordin:
## PR Overview
Background: ASM users are encountering issues submitting Classifications and we don't know why. One clue is that a number of said users tried using the Save Progress feature, and we detected that there may be an issue with `retrieveClassification()` function

This PR:
- Adds more console logs (for both successes and failures) which will be picked up by Rollbar, which should help future debugging.
- Fixes `retrieveClassification()` so that it no longer uses the `Array.shift()` method, because that actually changes the value of the `classification` object... and the Panoptes `classification` object should _only_ have its values changed via the `classification.update()` method.
  - Note: As far as I can determine, the values shift()ed in the `classification()` _don't need to be changed_ at retrieveClassification().
  - i.e. `retrievedClassification.links.subjects` should remain as is (shift() would turn it into an empty array) and `retrievedClassification.annotations` get overwritten later anyway with proper `.update()` calls.

### Status
@wgranger This is ready for review, but may require some pretty thorough testing.

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -236,10 +236,11 @@ const retrieveClassification = (id) => {
     
     apiClient.type('classifications/incomplete').get({ id })
       .then(([classification]) => {
-      
-        //TODO: Test if classification.annotations.shift() is OK; normally we don't update the classification object directly.
-        const subjectId = classification.links.subjects.shift();
-        const annotations = classification.annotations.shift();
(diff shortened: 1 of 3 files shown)


wgranger:
This branch looks good. Checks appear to be sound and strengthens the code, although we should continue taking a look at these ducks and how we can improve the error messaging.

PR #240, opened 2018-01-26
Remove Geordi and set maxItems

wgranger:
In Order to reduce the amount of errors and noise sent to Rollbar, we are setting `maxItems` to ten, as we've decided this is enough to deduce problems arising without getting unnecessary errors repeating hundreds of times.

Also, Geordi is removed as GA seems to provide the analytics we need. Currently, Geordi is reporting many superfluous errors.


src/components/App.jsx
```diff
@@ -32,7 +31,7 @@ class App extends React.Component {
     }
   }
 
-  returnSomething(something) { // eslint-disable-line class-methods-use-this
+  returnSometrolhing(something) { // eslint-disable-line class-methods-use-this
```
shaunanoordin: Uh, what even is this? I mean, it doesn't matter, but whuuh?

PR #243, opened 2018-01-26
Add Finally Polyfill

wgranger:
I have a feeling we might need this finally polyfill, and we may even want to consider another solution since the [Polyfill service](https://polyfill.io/v2/docs/features/) doesn't support Edge 7-11

### DIFF
--- a/src/index.tpl.html
+++ b/src/index.tpl.html
@@ -36,6 +36,6 @@
     <div id="root"></div>
     <%= htmlWebpackPlugin.options.gtm %>
 
-    <script src="https://ft-polyfill-service.herokuapp.com/v2/polyfill.min.js?features=default,es6,Array.prototype.includes,Array.prototype.keys"></script>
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Tests OK on Win7+IE11 and Win10+Edge, btw.

Small update:

@wgranger The reason the Polyfill service table _looks_ like `Promise.prototype.finally` _doesn't_ take care of Edge 7-11 is because...

1. "Edge 7, 8, 9, 10, 11" is actually Internet Explorer 7, 8, 9, 10, 11. ("Internet Explorer 12 onwards is actually Edge. Same thing, new branding.)
2. The `Promise` polyfill (which targets IE 7-11, aka Edge 7-11) should _already_ have `Promise.prototype.finally` as part of the package, hence there should have been no need to specify this subset.

screen shot 2018-01-29 at 18 27 24

So this means the addition of `Promise.prototype.finally` in this PR will more likely help Edge users than I11 users. That said, given the amount of errors we're seeing in Rollbar, heck, I'll take whatever fixes we can get.

PR #244, opened 2018-01-29
Failed Classifications with 422 errors shouldn't pop an alert

shaunanoordin:
## PR Overview
This is a messaging issue. When a Classification is submitted and a 422 is received in response, it means the Classification is invalid (something must have gone horribly wrong with the code) and therefore _cannot_ be re-queued.

This PR ensures that the "we'll try again later" message _only_ appears when a failed Classification is re-queued. (i.e. anything but a 422)

### Status
Ready for review, @wgranger

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -172,8 +172,6 @@ const saveAllQueuedClassifications = (dispatch, user = null) => {
       //FAILURE
       .catch((err) => {
         //Ah, crap.
-        itemsFailed++;
-        
         console.error('ducks/classifications.js saveAllQueuedClassifications() error: ', err);
         Rollbar && Rollbar.error &&
         Rollbar.error('ducks/classifications.js saveAllQueuedClassifications() error: ', err);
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Background for this PR, from today's Slack convo:

Will:
> We need to figure out a solution to a case where a user has an item stuck in their queue that constantly fails while newer items are submitted successfully. In this instance, an alert will appear even if newer items are submitted successfully.

Shaun:
> Question is, why would the older items fail?
> Say we have Classification A (which failed) then user does Classification B (which succeeds)... then A is requeued but _fails again)
> If it's a 422, then A is just an invalid classification and an alert shouldn't even appear. (Action to take: remove `failedItems++` in the case of a 422)
> If it's anything else, e.g. network down... then it doesn't make sense that B succeeds on its first try, but A fails on its retry. (It's a possible edge case with some very wonky network traffic)

wgranger:
LGTM thanks for putting this up

PR #245, opened 2018-01-30
Replace Promise.prototype.finally()

shaunanoordin:
## PR Overview
Some browsers, for some unknown reason, can't properly handle `Promise.prototype.finally()`. This PR adds a workaround where there's still a way to run a "account for Classification.save() succeeding or failing" function after a `Classification.save()`.

This error was reported in two ways:
- Users comment that after they successfully succeeded in sending a Classification, they couldn't get the next Subject. (This is consistent with .finally() not running)
- There were specific `.finally()` errors reported by Rollbar, right before it died. Skynet rest its digital soul.

### Status
@wgranger this is ready for a review and, if it's OK, a deploy.

### DIFF
--- a/src/ducks/classifications.js
+++ b/src/ducks/classifications.js
@@ -148,6 +148,36 @@ const saveAllQueuedClassifications = (dispatch, user = null) => {
     let itemsFailed = 0;
     const itemsToProcess = queue.length;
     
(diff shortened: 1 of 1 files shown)


wgranger:
LGTM. Similar code without the finally(), which is exactly what we wanted.

PR #246, opened 2018-01-30
Change Content and Minor Errors

### LINKED ISSUE #239: Change "Annotate" Button to "Transcribe"
- [x] Change button name in Toolbar

Mentions of Annotate button to check/change:
- [x] Directions

About page:
- [x] Significance of the Correspondence > 2nd paragraph: please change "blacks" to "Black people"

wgranger:
This PR covers several items:

- Tweaks content, including typos
- Adjusts some problems with the timeline (the long gray event covering up another event, having to double click to activate an event)
- Adds note on landing page about mobile devices and tablets
- Solves some linter errors as they appear
- Fixes some existence issues on several components that appeared via Rollbar
- Adds links to Subjects of Note on landing page
- Removes the `finally()` polyfill, which is no longer needed

Closes #239

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -52,12 +52,9 @@ class SelectedAnnotation extends React.Component {
 
   componentWillUnmount() {
     document.removeEventListener('keyup', this.handleKeyUp);
-    this.inputText.removeEventListener('mousedown', () => {
-      this.dialog.className = DISABLE_DRAG;
-    });
-    this.inputText.removeEventListener('mouseup', () => {
-      this.dialog.className = ENABLE_DRAG;
-    });
+    if (this.inputText) {
+      this.removeEventListeners();
+    }
   }
 
   toggleShowAnnotations() {
@@ -323,11 +320,22 @@ class SelectedAnnotation extends React.Component {
     this.props.onClose();  //Note that deleteSelectedAnnotation() also runs unselectAnnotation(), but this needs to be called anyway to inform the parent component.
   }
 
+  removeEventListeners() {
+    this.inputText.removeEventListener('mouseup', () => {
+      this.dialog.className = ENABLE_DRAG;
+    });
+    this.inputText.removeEventListener('mousedown', () => {
+      this.dialog.className = DISABLE_DRAG;
+    });
+  }
+
   handleKeyUp(e) {
     if (Utility.getKeyCode(e) === KEY_CODES.ESCAPE) {
+      this.removeEventListeners();
       this.cancelAnnotation();
     }
     if (Utility.getKeyCode(e) === KEY_CODES.ENTER && e.ctrlKey) {
+      this.removeEventListeners();
       this.saveText();
     }
   }
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -10,8 +10,8 @@ const PANE_WIDTH = 800;
 const PANE_HEIGHT = 350;
 const BUFFER = 10;
 
-const ENABLE_DRAG = "handle selected-annotation";
-const DISABLE_DRAG = "selected-annotation";
+const ENABLE_DRAG = 'handle selected-annotation';
+const DISABLE_DRAG = 'selected-annotation';
 
 class SelectedAnnotation extends React.Component {
   constructor(props) {
@@ -41,27 +41,18 @@ class SelectedAnnotation extends React.Component {
     }
     document.addEventListener('keyup', this.handleKeyUp);
     this.setState({ annotationText: text });
-    this.inputText.addEventListener('mousedown', () => {
-      this.dialog.className = DISABLE_DRAG;
-    });
-    this.inputText.addEventListener('mouseup', () => {
-      this.dialog.className = ENABLE_DRAG;
-    });
     this.inputText.focus();
   }
 
   componentWillUnmount() {
     document.removeEventListener('keyup', this.handleKeyUp);
-    if (this.inputText) {
-      this.removeEventListeners();
-    }
   }
 
   toggleShowAnnotations() {
     if (this.context.googleLogger) {
       this.context.googleLogger.logEvent({
         type: 'click-dropdown',
-        timestamp: new Date().toISOString()
+        timestamp: new Date().toISOString(),
       });
     }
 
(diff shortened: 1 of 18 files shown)


src/components/SelectedAnnotation.jsx
```diff
@@ -52,12 +52,9 @@ class SelectedAnnotation extends React.Component {
 
   componentWillUnmount() {
     document.removeEventListener('keyup', this.handleKeyUp);
-    this.inputText.removeEventListener('mousedown', () => {
-      this.dialog.className = DISABLE_DRAG;
-    });
-    this.inputText.removeEventListener('mouseup', () => {
-      this.dialog.className = ENABLE_DRAG;
-    });
+    if (this.inputText) {
```
shaunanoordin: I don't understand why the `if ()` existential check is _outside_ `removeEventListeners()`

Logically, I'd imagine it to be something like...

```
removeEventListeners() {
  if (this.inputText) {
    this.inputText.removeEventListener('mouseup', this.handleMouseUp);
    this.inputText.removeEventListener('mousedown', this.handleMouseDown);
  }
}
```

...which ensures that `removeEventListeners()` can be called willy-nilly and it'd still be safe to do so.

src/components/SelectedAnnotation.jsx
```diff
@@ -323,11 +320,22 @@ class SelectedAnnotation extends React.Component {
     this.props.onClose();  //Note that deleteSelectedAnnotation() also runs unselectAnnotation(), but this needs to be called anyway to inform the parent component.
   }
 
+  removeEventListeners() {
+    this.inputText.removeEventListener('mouseup', () => {
```
shaunanoordin: Wait, I just realised this doesn't work whatsoever - even our current code is wrong.

When calling...
```
htmlElement.addEventListener(eventName, eventHandler);
htmlElement.removeEventListener(eventName, eventHandler);
```
... `eventHandler` has to be the _same thing._

So you can have, say...
```
const eventName = 'mouseup';
const eventHandler = this.handleMouseUp;

htmlElement.addEventListener(eventName, eventHandler);
htmlElement.removeEventListener(eventName, eventHandler);
```
...and it'll work fine.

However, calling...
```
htmlElement.addEventListener(eventName, () => { console.log('This is anonymous handler A') });
htmlElement.removeEventListener(eventName, () => { console.log('This is anonymous handler B') });
```
...means that we _keep adding_ (anonymous) event listeners to `htmlElement` and never remove any.

I suspect this hasn't been an issue for us because...
- when the React component unmounts, it annihilates the `htmlElement` itself, so it doesn't matter if it was overloaded with six hundred anonymous event handlers.
- in the case of `document.addEventListener`, we properly use a common reference event handler (`this.handleKeyUp`)

Recommendations:
- Create `SelectedAnnotation.handleKeyUp` and `SelectedAnnotation.handleKeyDown` methods.
- Make sure all `this.whatever.removeEventListener()` calls reference those methods instead of generating new anonymous ones.

src/components/SelectedAnnotation.jsx
```diff
@@ -323,11 +320,22 @@ class SelectedAnnotation extends React.Component {
...
+      this.dialog.className = ENABLE_DRAG;
+    });
+    this.inputText.removeEventListener('mousedown', () => {
+      this.dialog.className = DISABLE_DRAG;
+    });
+  }
+
   handleKeyUp(e) {
     if (Utility.getKeyCode(e) === KEY_CODES.ESCAPE) {
+      this.removeEventListeners();
```
shaunanoordin: IIRC, you said that we had to remove the event listeners _before_ `componentWillUnmount` runs, because there's a bizarre situation where HTML elements are destroyed (in this case, `this.input`) _before_ the unmount process begins (instead of after it finishes, as expected), correct?

Did we figure out why? I'm still flabbergasted by this React WTF-ery.

shaunanoordin: If it's confirmed that the issue is React WTF-ery, (and therefore we can't rely on `componentWillUnmount`) then we'll need to consider that it may be insufficient to just have `this.removeAllEventListener()` at the ESC call.

If we're going down this path, it would make more sense to
1. call `this.removeAllEventListeners()` before every instance of `this.props.onClose()` instead. (This would target _all_ user actions that gets rid of the transcription box)
2. fold `document.removeEventListener()` into `this.removeAllEventListener()`, since we're essentially saying that `this.removeAllEventListeners` is _replacing_ `componentWillUnmount()`

src/components/SelectedAnnotation.jsx
eatyourgreens: React destroys any event listeners that it created from component props, on unmount. If I remember the React docs correctly, it doesn't remove any listeners that you added with `addEventListener`.

[React says it supports `onMouseUp` now](https://reactjs.org/docs/events.html#mouse-events), so you might not have to manually add these listeners yourselves.

src/containers/SubjectViewer.jsx
```diff
@@ -300,7 +300,7 @@ class SubjectViewer extends React.Component {
   /*  Once the Subject has been loaded properly, fit it into the SVG Viewer.
    */
   onImageLoad() {
-    if (this.svgImage.image) {
+    if (this.svgImage && this.svgImage.image) {
```
shaunanoordin: 👍

shaunanoordin:
Jim's comments makes perfect sense. We initially used the `item.addEventListener` pattern for global events (usually key presses, when we're unsure where the browser focus is) and the SVG elements (which IIRC required very specific interaction handling), but I think we accidentally followed that pattern for SelectedAnnotation instead of the standard React pattern.

i.e. for `SelectedAnnotation` it makes more sense to simply rewrite the code as...
```
render() {
  return (
    ...
    <input
      className={inputClass}
      type="text" ref={(c) => { this.inputText = c; }}
      onChange={this.onTextUpdate} value={this.state.annotationText}

      //Replace addEventListeners/removeEventListeners with these
      onMouseDown={() => { this.dialog.className = DISABLE_DRAG }}
      onMouseUp={() => { this.dialog.className = ENABLE_DRAG }}

    />
    ...
  );
}
```

For consistency, this standardisation should be applied across:
- `components/SelectedAnnotation`
- `components/SaveClip`
- `components/Navigator` ❓ 
  - (Note 1: Not sure if standardisation works here due to the SVG element)
  - (Note 2: in any case, this is actually missing removeEventListener)

Side note: `components/Crop` also does something specific with event handling, but that can be double-checked separately.

src/components/SelectedAnnotation.jsx
wgranger: I think the elements are removed before the unmount process begins because the `onClose` function runs and removes the dialog box before the child (`SelectedAnnotation`) has the opportunity to run the remove listener cleanup.

shaunanoordin:
## PR Update
Most recent commit addresses issue of the how components handle events. So now we expect no further "can't unmount X" errors from users.

### Status
Nice one, @wgranger! Approved, merging, go.

PR #251, opened 2018-02-13
Update Panoptes Client, handle timeout sessions

shaunanoordin:
## PR Overview
This PR aims to solve the issue where a user's login token may expire while the user is in the process of transcribing some work, and the Panoptes Client forgets to refresh the token.

- The Panoptes Client has been updated to 2.9.2, which now includes a `apiClient.beforeEveryRequest()` hook and an `oauth.checkBearerToken()` function.
- See zooniverse/panoptes-javascript-client#81 for details on how such unexpected timeouts might occur.

### Implementation notes
Jim provided some useful notes:
```
apiClient.beforeEveryRequest = function() {
  return oauth.checkBearerToken()
  .then(function (token) {
    // still logged in. Probably don't need to do anything here
  })
  .catch(function (error) {
    // Token couldn't be renewed. Message that they are about to be logged out
    // or call oauth.signIn(url) to redirect to panoptes.zooniverse.org and get a new session.
})
```
> Jim:
> I guess you’d need to be careful to save any work in progress if you need to redirect to zooniverse.org to get a new session.

### Status
WIP, that "save progress before reloading" work needs to be implemented.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -62,7 +62,7 @@ class App extends React.Component {
         if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {          
           this.props.dispatch(emergencySaveWorkInProgress());
           this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
-          return Promise.reject('User is supposed to be logged in, but token has expired.');  //WARNING: Testing has show that this does NOT abort the Panoptes request.
+          return Promise.reject(new Error('User is supposed to be logged in, but token has expired.'));
           //The intent is that if the user is supposed to be logged in but
           //isn't, the whole request (that comes after .beforeEveryRequest)
           //should not continue.
@@ -71,6 +71,7 @@ class App extends React.Component {
       })
       .catch((err) => {
         console.error('App.checkIfLoggedInUserIsStillLoggedIn() error: ', err);
+        return Promise.reject(err);  //The Promise.reject must be passed on to stop the parent API request.
       });
   }
 
@@ -142,15 +143,18 @@ class App extends React.Component {
           //DEBUG/TEST
           //----------------
           //Will: enable this for testing. Once PR is merged, I'll have a separate PR to delete the debugs. (@shaun 20180223)
-          /*
-          (env !== 'staging' && env !== 'development') ? null :
+          //Will: enable this for testing. Once PR is merged, I'll have a separate PR to delete the debugs. (@shaun 20180223)
+          /*(env !== 'staging' && env !== 'development') ? null :
           <div style={{position: 'fixed', top: '0', left: '0'}}>
             <button onClick={()=>{
-              apiClient.type('me').get().then((data)=>{ console.log(data); });
+              apiClient.type('me').get().then((data)=>{ console.log('/me : ', data); });
             }}>TEST: FETCH USER DETAILS</button>
             <button onClick={()=>{
               localStorage.setItem('TEST_TOKEN_ERROR', 'true');
             }}>TEST: FAKE ERROR</button>
+            <button onClick={()=>{
+              localStorage.removeItem('TEST_TOKEN_ERROR');
+            }}>TEST: !FAKE ERROR</button>
             <button onClick={()=>{
               this.props.dispatch(emergencySaveWorkInProgress());
             }}>TEST: SAVE WIP</button>
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -8,11 +8,13 @@ import { toggleDialog } from '../ducks/dialog';
 
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
+import DialogOfFailure from './DialogOfFailure';
 import Dialog from './Dialog';
 import LoadingSpinner from './LoadingSpinner';
 import { generateSessionID } from '../lib/get-session-id';
 
 import apiClient from 'panoptes-client/lib/api-client';
+import auth from 'panoptes-client/lib/auth';
 import oauth from 'panoptes-client/lib/oauth';
 
 import { env } from '../config';
@@ -50,18 +52,19 @@ class App extends React.Component {
       .then((token) => {
         console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
       
+        //DEBUG/TEST
+        //----------------
         const DEBUG_TEST = localStorage.getItem("TEST_TOKEN_ERROR");  //Use localStorage.TEST_TOKEN_ERROR = true to trigger the dialog
-      
+        //----------------
+
         //If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
-        if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {
-          
-          this.props.dispatch(toggleDialog(<div>Oh no</div>, false, true));
-          //TODO
-          
-          return false;
+        if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {          
+          this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
+          //TODO          
+          return Promise.reject('User is supposed to be logged in, but token has expired.');
         }
       
-        return true;
+        return;
       })
       .catch((err) => {
         console.error('App.checkIfLoggedInUserIsStillLoggedIn() error: ', err);
@@ -130,6 +133,21 @@ class App extends React.Component {
           <Dialog>
             {this.props.dialog}
           </Dialog>
+        }
+        
+        {
+          //DEBUG/TEST
+          //----------------
+          (env !== 'staging' && env !== 'development') ? null :
+          <div style={{position: 'fixed', top: '0', left: '0'}}>
+            <button onClick={()=>{
+              apiClient.type('me').get().then((data)=>{ console.log(data); });
+            }}>TEST: FETCH USER DETAILS</button>
+            <button onClick={()=>{
+              localStorage.setItem('TEST_TOKEN_ERROR', 'true');
+            }}>TEST: DELETE TOKEN</button>
+          </div>
+          //----------------
         }
       </div>
     );
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -142,6 +142,8 @@ class App extends React.Component {
         {
           //DEBUG/TEST
           //----------------
+          //Will: enable this for testing. Once PR is merged, I'll have a separate PR to delete the debugs. (@shaun 20180223)
+          /*
           (env !== 'staging' && env !== 'development') ? null :
           <div style={{position: 'fixed', top: '0', left: '0'}}>
             <button onClick={()=>{
@@ -156,7 +158,7 @@ class App extends React.Component {
             <button onClick={()=>{
               this.props.dispatch(emergencyLoadWorkInProgress());
             }}>TEST: LOAD WIP</button>
-          </div>
+          </div>*/
           //----------------
         }
       </div>
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -10,8 +10,8 @@ import { emergencySaveWorkInProgress, emergencyLoadWorkInProgress } from '../duc
 
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
-import DialogOfFailure from './DialogOfFailure';
 import Dialog from './Dialog';
+import DialogOfFailure from './DialogOfFailure';
 import LoadingSpinner from './LoadingSpinner';
 import { generateSessionID } from '../lib/get-session-id';
 
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -6,7 +6,6 @@ import { fetchProject, PROJECT_STATUS } from '../ducks/project';
 import { disableBanner } from '../ducks/banner';
 import { toggleDialog } from '../ducks/dialog';
 
-//DEBUG/TEST
 import { emergencySaveWorkInProgress, emergencyLoadWorkInProgress } from '../ducks/emergency-save';
 
 import Header from './Header';
@@ -62,11 +61,10 @@ class App extends React.Component {
 
         //If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
         if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {          
+          this.props.dispatch(emergencySaveWorkInProgress());
           this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
-          //TODO          
           return Promise.reject('User is supposed to be logged in, but token has expired.');
-        }
-      
+        }      
         return;
       })
       .catch((err) => {
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -52,10 +52,9 @@ class App extends React.Component {
     
     return oauth.checkBearerToken()
       .then((token) => {
-        console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
-      
         //DEBUG/TEST
         //----------------
+        console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
         const DEBUG_TEST = localStorage.getItem("TEST_TOKEN_ERROR");  //Use localStorage.TEST_TOKEN_ERROR = true to trigger the dialog
         //----------------
 
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -68,11 +68,8 @@ class App extends React.Component {
           //should not continue.
         }      
         return;
-      })
-      .catch((err) => {
-        console.error('App.checkIfLoggedInUserIsStillLoggedIn() error: ', err);
-        return Promise.reject(err);  //The Promise.reject must be passed on to stop the parent API request.
       });
+      //Note: do NOT add a catch() here, or else Promise.reject() above won't stop the API request.
   }
 
   getChildContext() {
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -4,6 +4,8 @@ import { connect } from 'react-redux';
 import { ZooFooter } from 'zooniverse-react-components';
 import { fetchProject, PROJECT_STATUS } from '../ducks/project';
 import { disableBanner } from '../ducks/banner';
+import { toggleDialog } from '../ducks/dialog';
+
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
 import Dialog from './Dialog';
@@ -32,23 +34,32 @@ class App extends React.Component {
       props.dispatch(checkLoginUser());
     }
     
-    //TODO: This is just a test.
-    apiClient.beforeEveryRequest = () => {
-      return oauth.checkBearerToken()
-        .then((token) => {
-          //All is OK
-          console.log('x'.repeat(100), '\nALL IS OK');
-        })
-        .catch((error) => {
-          //Token has expired! How? Why? We need to force a refresh.
-          console.error('x'.repeat(100), '\nERROR, ERROR');
-        });
-    };
-    //--------------------------------
+    //SPECIAL: users on ASM tend to stay in a single session for WAY longer
+    //than a standard Zooniverse CFE user, so we're encountering issues where
+    //their login tokens timeout in the middle of classifying a document.
+    //The following is a mechanism that checks if a user is logged in before
+    //performing any action that may require a valid login token, taking safety
+    //actions if necessary.
+    apiClient.beforeEveryRequest = this.checkIfLoggedInUserIsStillLoggedIn.bind(this);
   }
-
-  returnSometrolhing(something) { // eslint-disable-line class-methods-use-this
-    return something;
+  
+  checkIfLoggedInUserIsStillLoggedIn() {
+    const props = this.props;
+    
+    return oauth.checkBearerToken()
+      .then((token) => {
+        console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
+      
+        this.props.dispatch(toggleDialog(<div>Oh no</div>, false, true));
+      
+        //If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
+        if (props.initialised && props.user && !token) {
+          //TODO
+        }
+      })
+      .catch((err) => {
+        console.error('App.checkIfLoggedInUserIsStillLoggedIn() error: ', err);
+      });
   }
 
   getChildContext() {
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -6,6 +6,9 @@ import { fetchProject, PROJECT_STATUS } from '../ducks/project';
 import { disableBanner } from '../ducks/banner';
 import { toggleDialog } from '../ducks/dialog';
 
+//DEBUG/TEST
+import { emergencySaveWorkInProgress, emergencyLoadWorkInProgress } from '../ducks/emergency-save';
+
 import Header from './Header';
 import ProjectHeader from './ProjectHeader';
 import DialogOfFailure from './DialogOfFailure';
@@ -145,7 +148,13 @@ class App extends React.Component {
             }}>TEST: FETCH USER DETAILS</button>
             <button onClick={()=>{
               localStorage.setItem('TEST_TOKEN_ERROR', 'true');
-            }}>TEST: DELETE TOKEN</button>
+            }}>TEST: FAKE ERROR</button>
+            <button onClick={()=>{
+              this.props.dispatch(emergencySaveWorkInProgress());
+            }}>TEST: SAVE WIP</button>
+            <button onClick={()=>{
+              this.props.dispatch(emergencyLoadWorkInProgress());
+            }}>TEST: LOAD WIP</button>
           </div>
           //----------------
         }
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -63,7 +63,10 @@ class App extends React.Component {
         if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {          
           this.props.dispatch(emergencySaveWorkInProgress());
           this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
-          return Promise.reject('User is supposed to be logged in, but token has expired.');
+          return Promise.reject('User is supposed to be logged in, but token has expired.');  //WARNING: Testing has show that this does NOT abort the Panoptes request.
+          //The intent is that if the user is supposed to be logged in but
+          //isn't, the whole request (that comes after .beforeEveryRequest)
+          //should not continue.
         }      
         return;
       })
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -50,12 +50,18 @@ class App extends React.Component {
       .then((token) => {
         console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
       
-        this.props.dispatch(toggleDialog(<div>Oh no</div>, false, true));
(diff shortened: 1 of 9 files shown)


shaunanoordin:
```
apiClient.beforeEveryRequest = () => {
  return oauth.checkBearerToken()
    .then((token) => {
      //OUTCOME A
    })
    .catch((error) => {
      //OUTCOME B
  });
};
```

So here's an observation, due to the way that the Panoptes Client is coded:
- when everything is fine (i.e. no expired sessions), and the user is logged in, we expect to see Outcome A, with `token` being an object.
- when everything is fine, and the user is NOT logged in, we expect to see Outcome A, with `token` being null.
- when everything ISN'T fine (i.e. session has expired or is somehow invalid) and the user is logged in, we expect to see Outcome B, which will prompt us to reload.
- ❓ However... it appears that for a non-logged in user, when everything ISN'T fine, it's very likely that we'll see Outcome A instead of Outcome B.

This means that there's no way for this solution to work for a non-logged in user, as we can't tell when a non-logged in user's session has expired.

However, this SHOULDN'T matter since a non-logged in user shouldn't have a damn session to begin with. They can't use Save Progress, they can only use Submit Classification, which is considered a public feature and shouldn't require sessions.

However however, people HAVE been reporting Submit Classification giving them problems, and all debug data point to the sessions expiring.

However however however, those people are logged in users, and often we see that their problem with Submit Classification errors come _after_ they encounter errors with Save Progress (which does require a valid login session) so it's likely that the "Submit Classification" error is actually just one of the results of a cascade of errors originating from a faulty Save Progress operation.

In any case, I have a simple test set up to see what happens to a logged in user after they put their computer to sleep for a few hours.

shaunanoordin:
I'm back with updates, and the update is :sad_panda:

Panda

Testing:
- I logged into my computer at 12:15pm, awakening it from its deathless slumber. _Iä! Iä! Macbook fhtagn! Ph'nglui mglw'nfah Macbook Zooniverse wgah'nagl fhtagn!_
- I see the Chrome Incognito window that I left open since last night, logged in with my account. It should very very definitely have a logged-out account now
- Last action at 5:36pm yesterday, with status: `ALL IS OK, {access_token: "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzUxMiJ9.eyJkYXRhIjp7I…9YgKbyh6qjPlMaOeiHJDmL0Fez5Cj0VhwrtanJNqHgXx4sWpU", token_type: "bearer", expires_in: "7200", expires_at: 1518550563737}`
- New actions from this point on.
- Clicked on "Save Progress".
  - UI: no response. Expected "confirmation"
  - Console: `ALL IS OK, undefined`
  - Network: `https://panoptes-staging.zooniverse.org/api/classifications/90542?http_cache=true` returns a 404
- Clicked on "Save Progress" again.
  - UI: no response. Expected "confirmation"
  - Console: `ALL IS OK, null` (Wait, why did this change?)
  - Network: `https://panoptes-staging.zooniverse.org/api/classifications/90542?http_cache=true` returns a 404
- Repeat x2
- Clicked on "Finish"
  - UI: received error message. (Alert: "Your Classification couldn't be submitted")
  - Console: `ALL IS OK, null`
  - Network: `https://panoptes-staging.zooniverse.org/api/classifications/90542?http_cache=true` returns a 404 (Wait, what happened to a simple /classifications post)

eatyourgreens:
I think `/classifications/:id` is an authenticated endpoint. You have to be the owner of that classification in order to update it. @camallen can confirm whether the response should be 404 or 401.

Odd, and unexpected, that client resolves the promise with `null` rather than throwing an error when it tries to get the session. Do you show up as signed-in in the user interface too? Maybe you are signed-out, but still accessing the saved classification queue for the logged-in user.

shaunanoordin:
Quick notes before the thoughts evaporate from my brain:
- The earlier code block is insufficient because once I re-logged into my computer, the Panoptes client sees me as a non-logged in user (which as noted, always sees Outcome A), not a logged-in user whose session is expired and needs to be re-logged in (which leads to Outcome B, which I was hoping to trigger).
- Essentially, the Panoptes client can tell if a user is logged in or not, but can't tell if the user _SHOULD_ be logged in.
- Only the app (in this case, ASM) can tell if the user _should_ be logged in or not.
- Hence, we need the ability to use a discrepancy check, to compare what the app thinks is the login state of the user, and what the panoptes client says is the actual login state of the user. e.g.:

```
apiClient.beforeEveryRequest = () => {
  if (redux.login.initialised && redux.login.user) {  //The app thinks the user is logged in (i.e. user SHOULD be logged in)
    return oauth.checkBearerToken()
      .then((token) => {
        if (token) {  //User SHOULD be logged in and IS.
           //OUTCOME A
        } else {  //User SHOULD be logged in but ISN'T
          //OUTCOME C
        }
      })
      .catch((error) => {  //Something else went horribly wrong, like a network error.
        //OUTCOME B
    });
  } else {
    //I probably should think of something here but I'm not sure what
  }
};
```

OK, I'll continue documenting later. NOW IS NOM TIME.

eatyourgreens:
Quick question: how did you even manage to click on Save Progress if you weren't logged in?

eatyourgreens:
I’m an idiot. Errors in the auth flow are already caught in the client code, so `checkBearerToken` should be expected to resolve with a token, if we are still logged in, or null. `undefined` seems to be an error in the client, because the catch block doesn’t return a value. 

You’re right, it’s the responsibility of ASM, in this case, to know whether or not it was expecting a token, and act accordingly. 

Also, the client clears your auth headers just before your current token expires, preventing you from saving work until it gets a new token. This is a bug. 🐛

shaunanoordin:
Ah! I forgot to reply to your comments, Jim. OK, so:

1. Q: _"how did you even manage to click on Save Progress if you weren't logged in?"_ A: the app state still thinks the user is logged in so shows the Save Progress button. :sad_panda:

2. Re: _"client clears your auth headers just before your current token expires, preventing you from saving work until it gets a new token. This is a bug."_ Aye, but a bug that's caught - I see it's taken care of in zooniverse/panoptes-javascript-client#82 ! jim++

3. On that note, I need to see if I should update to the latest PJC 2.9.3 to see if that fixes the issue I reported in zooniverse/panoptes-javascript-client#84 . Either way, it shouldn't affect this PR, as long as I'm guaranteed that token === null (or even undefined) if the user ain't logged in.

Sorry if my replies are incomplete, I'm feeling a bit scatterbrained at the moment as I just did a chunk of updates tested against an absurd amount of scenarios and I need to recall everything I did for the PR update documentation. 😵

shaunanoordin:
## PR Update
Very simply, with this PR:
- the app now saves any transcription-work-in-progress to localStorage if the user encounters a session error.
  - A session error, in this case, is usually caused by the user's login session/token expiring in the middle of transcribing a document. This screws up their ability to save their work in progress and/or submit a Classification.
  - A session error is detected by CHECKING BEFORE EACH PANOPTES REQUEST if the user is _supposed_ to be logged in (App.props.user !== null) and comparing it to whether the user is _actually_ logged in (oauth.token !== null)
- When the user reloads the Classify page, if an emergency save is detected, it's restored.
  - The user has the option to start anew, if they'd rather ignore their old work in progress. (This is also a safety valve in case, for whatever reason, the emergency save data is never deleted from local storage.)
  - Actually, the order is: if there's a manual Save Progress detected, prompt the user. If not, check for an emergency safe. If not, grab the Subject queue as per usual. (Manual save trumps emergency save!)
- Any emergency save data is wiped after a Classification is placed in the Classification queue and is considered "safe enough". (Warning: be careful of failsafes interfering with other failsafes)
  - Important: we do NOT clear the emergency save after a Subject is loaded (e.g. by placing it in prepareForNewSubject()).
    - I found that we need to retain the emergency save until we're sure it's been submitted, because consider this scenario: 1. user is logged in and transcribes half a page. 2. user goes out for lunch. 3. user returns to transcribe rest of page, but session error has occurred (user was logged out in the middle of work). 4. app saves data and prompts user to reload the page. 5. user reloads the page; the emergency save is loaded, the Subject-in-progress is reinstated, and prepareForNewSubject() kicks in. 6. ❗️If the user is logged out, and they will be, they'll be _prompted to login again by going to an external website._ 7. The user returns to the app, and _once again:_ the emergency save is loaded, the Subject-in-progress is reinstated, and prepareForNewSubject() kicks in.
    - If we cleared the the emergency save at step 5, then when the user actually returns to continue their work at Step 7, all their emergency-save data would be gone.

### Notes
The emergency save is a blunt instrument acting as a safety net. It does not, for example, care if you're logged in to the same account when it restores your emergency save. So for example, if you transcribed half the page as `zootester1`, then encountered a session error, reloaded, was prompted to login, then logged in as `zootester2` (or continue as a non-logged-in user), and continued, it will not care and just load the saved progress anyway.

### Status
Ready for testing. GOOD LUCK WITH TESTING.

### Testing Madness
- A non-logged-in user should be able to make an annotation and submit.
  - And do it again. (i.e. testing the queue)
- A logged-in user user should be able to make an annotation and submit.
  - And do it again. (i.e. testing the queue)
- A logged-in user should be able to make an annotation, Save Progress, reload, _continue their work,_ and submit.
- A logged-in user should be able to make an annotation, Save Progress, reload, _start on a new document instead,_ and submit.
- 💫 A logged-in user should be able to make an annotation, encounter a session error, reload the page, (and be prompted to login again, which would reload the page _again_), be prompted to continue their work, and submit.
  - And after that, make an annotation on a new subject and submit. (i.e. testing that the emergency safe system deactivates for the next Subject, so that user isn't forever stuck annotating that one document.)
  - 💀 PREFERABLY, the next Subject should not be the same Subject that the user just continued work on, but this may be unavoidable if the emergency-saved Subject also appears in the fresh Subject queue.
- As 💫 above, except when prompted to continue their work, user can choose to start anew, then submit.
- 💥 A logged-in user should be able to make an annotation, Save Progress, encounter a session error, reload the page, and be prompted to continue _from their manually saved progress and not the emergency saved progress,_ then submit.
  - And do it again. (i.e. testing the queue)

There are two ways to trigger a session error for testing purposes:
1. modify your local `node_modules/panoptes-client` to reduce the lifetime of an access token to tiny tiny amounts.
2. Enable the `//DEBUG/TEST` code in App.jsx and jury-rig the session error simulators (that's a whole different documentation manual.)

NOTE: I will clean up the `//DEBUG/TEST` code in a separate PR, after this PR has been OK-ed.

shaunanoordin:
Me, looking at the hotfixes needed for the ASM login bugs: ["This is fine."](https://giphy.com/gifs/fire-dont-care-unphased-l378m8mNCUdjA3Euk)

src/components/App.jsx
```diff
@@ -29,10 +37,42 @@ class App extends React.Component {
...
-
-  returnSometrolhing(something) { // eslint-disable-line class-methods-use-this
-    return something;
+  
+  checkIfLoggedInUserIsStillLoggedIn() {
+    const props = this.props;
+    
+    return oauth.checkBearerToken()
+      .then((token) => {
+        console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
```
wgranger: I think it would be good at some point, perhaps after this login issue is resolved, to remove the logs we have in the app that were initially used to pinpoint this bug.

shaunanoordin: Thanks for spotting this - this console.log (and any others I missed) should have been marked with that //DEBUG/TEST tag so I could Ctrl+F and remove 'em in the follow up cleanup.

wgranger:
Should emergency save trump the manual save? Perhaps this is a conversation for the channel, but it seems there might be an instance where someone has begun working for a while, hasn't manually saved for several lines, and then they eventually get the logout bug. They would then lose those several lines since their manual saved state is a bit behind. 

Is there a reason we would want to give priority to the manual, and potentially older, save of a classification in progress?

src/ducks/emergency-save.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+};
+
+const emergencyLoadWorkInProgress = () => {
+  return (dispatch, getState) => {
+    try {
+      const subjectId = localStorage.getItem(SUBJECT_ID_KEY);  //TODO: Check if a type conversion is required.
+      const annotations = JSON.parse(localStorage.getItem(ANNOTATIONS_KEY));
+      dispatch(fetchSavedSubject(subjectId));
+      prepareForNewSubject(dispatch, null, false);
```
wgranger: Why is this receiving three arguments (dispatch, null, false)? `prepareForNewSubject` expects two arguments, although I can understand the reason for passing `null`, as the subject's previous annotations will already be retrieved in the `fetchSavedSubject()` function.

shaunanoordin:
re: does manual save trump emergency save, or does emergency save trump manual save - yeah, this is up for discussion. My reasoning for manual save >= emergency save is that I always prefer giving control to the user where possible, but you make a good point about a user who, e.g. makes a manual save at the 5 minutes mark and then spends 1 hour not saving before encountering a bug.

I can switch the code to emergency save >= manual save by switching the order of if/else in...
```
ducks/subject.js fetchSubject()...
    if (savedClassificationPrompt) {  //Check if the user has manually saved progress
      dispatch(toggleDialog(<ClassificationPrompt />, false, true));
    } else if (checkEmergencySave()) {  //If not, check if there's an emergency save. (Manual save trumps emergency save.)
      dispatch(emergencyLoadWorkInProgress());
      dispatch(toggleDialog(<DialogOfContinuation />, false, false));
    } else if ...
```
...but I'll need to check with you re: how or if'd we'd delete the manual save. (Initial thoughts: add `clearManualSaveData(); clearEmergencySaveData();` at the end of `ducks/classification.js submitClassification()` since once a Classification is submitted we can assume that we want a funky fresh new subject)

src/ducks/emergency-save.js
```diff
@@ -0,0 +1,64 @@
...
+import { setAnnotations } from './annotations';
+
+const SUBJECT_ID_KEY = 'emergency_save_subjectId';
+const ANNOTATIONS_KEY = 'emergency_save_annotations';
+
+function checkEmergencySave() {
+  const subjectId = localStorage.getItem(SUBJECT_ID_KEY);
+  const annotations = localStorage.getItem(ANNOTATIONS_KEY);
+
+  return !!subjectId && !!annotations;
```
wgranger: Should there be a check in this function, or before, that a user is logged in? Ex. Lets say a user must refresh his/her page because the token is expired and they are logged out. When the page reloads (on the classifier), the `DialogOfContinuation` will appear allowing a user to load the saved annotation without necessarily logging on again.

wgranger: Ex. The `DialogOfFailure` says "you may need to reload the page and log in again," but the page can be reloaded and the emergency classification can be retrieved again without logging back in. 

Note: this is likely coming from line 250 of the subject duck. However, I can understand this is the functionality you want with a non-logged in user, but it might throw off the logged in user who begins a classification logged in and continues as a guest.

wgranger:
Note: I've been testing some of this branch by changing the token allowance in the node_modules oauth to var `TOKEN_EXPIRATION_ALLOWANCE = 119 * 60 * 1000;` to signify 119 minutes (giving on minute before the token would "expire."

src/ducks/emergency-save.js
shaunanoordin: Will, you are correct - the emergency save does NOT care if the user is logged in, and it's entirely possible for a non-logged-in User X to continue a logged-in User Y's work (if that User Y was booted off by an expired token). It's designed this way to favour the user experience, allowing them to see their work even if they 'forget' to re-login.

This was buried in my comments above, sorry:
https://github.com/zooniverse/anti-slavery-manuscripts/pull/251#issuecomment-368192732
> The emergency save is a blunt instrument acting as a safety net. It does not, for example, care if you're logged in to the same account when it restores your emergency save. So for example, if you transcribed half the page as zootester1, then encountered a session error, reloaded, was prompted to login, then logged in as zootester2 (or continue as a non-logged-in user), and continued, it will not care and just load the saved progress anyway.

shaunanoordin:
Will++ for the insanity that is testing this, especially with all the timeouts required.

On the TODO:
- [ ] Add Rollbar reporting for when emergencySave() triggers. (Will need to check if adding logging to emergencyLoad() is worth it.)
- [ ] I'll need to see if it's worth upgrading `panoptes-client` to 2.9.4 (recently released!)
- [ ] Switch priority: EMERGENCY SAVE should trump MANUAL SAVE. (As discussed on Slack, thanks to everyone weighing in on this.)

src/components/App.jsx
```diff
@@ -29,10 +37,41 @@ class App extends React.Component {
...
+        //----------------
+        console.log(`checkBearerToken - initialised: ${!!props.initialised}, user: ${!!props.user}, token: ${!!token}`);
+        const DEBUG_TEST = localStorage.getItem("TEST_TOKEN_ERROR");  //Use localStorage.TEST_TOKEN_ERROR = true to trigger the dialog
+        //----------------
+
+        //If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
+        if (!!DEBUG_TEST || (props.initialised && props.user && !token)) {          
+          this.props.dispatch(emergencySaveWorkInProgress());
+          this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
+          return Promise.reject('User is supposed to be logged in, but token has expired.');  //WARNING: Testing has show that this does NOT abort the Panoptes request.
```
eatyourgreens: [Promise.reject](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) on MDN recommends that the reason be an Error: `return Promise.reject(new Error('Token has expired'));`

I'm not sure why this isn't breaking the promise chain and aborting the API request.

shaunanoordin: Oh SNAAAP it's because the `.catch()` stops the chain from breaking!

```
checkIfLoggedInUserIsStillLoggedIn() {
  return oauth.checkBearerToken()
    .then((token) => {
      return Promise.reject(new Error('pls stop the rest of the API request'));
    })
    .catch((err) => {
      //DERP the Promise.reject from above gets caught here, and the rest of the Promise resolves normally.
    });
}
```

Correct way to kill that API request:
```
checkIfLoggedInUserIsStillLoggedIn() {
  return oauth.checkBearerToken()
    .then((token) => {
      return Promise.reject(new Error('pls stop the rest of the API request'));
    })
    .catch((err) => {
      return Promise.reject(err);  //THIS will absolutely definitely kill the request.
      //Mind you, you might get double error messages logged if the external API request also has console.error().
    });
}
```

Another way to prove that Promise.reject() actually kills the parent API request is to remove that .catch() protection.

eatyourgreens: You could also reverse the order of the try and catch blocks.
```javascript
return oauth.checkBearerToken()
  .catch(error => // catch any errors thrown by the API client)
  .then(token => // rejecting here will block the chain until the next catch block)
```

eatyourgreens: Thinking about it, do you really need to catch errors here? I guess the network might be down, or Panoptes might not be responding to requests, but I don't know if the OAuth module passes those errors back anyway.

@simoneduca this might also be relevant to the work you are doing in https://github.com/zooniverse/shakespeares_world/pull/369

shaunanoordin:
I'm going to repeat some really good points that @eatyourgreens made which got buried in the inline comments (now hidden because I just pushed code updates). Hopefully this makes it easier to see what the discussion was about. (Hey, @simoneduca ! )

If you want the **oauth.checkBearerToken function to terminate the API request** (that comes after _apiClient.beforeEveryRequest,_ ) - e.g., because you detected the user has lost their login session - make sure you use **Promise.reject()** and make sure it's _not_ caught by a catch().

Usable pattern (ver 1):
```
return oauth.checkBearerToken()
  .then((token) => {
    if (this.user && !token) { return Promise.reject(new Error('Cannot proceed, user is supposed to be logged in but has no valid access token')) }  //This stops the rest of the API request.
  });
  //Do NOT add a catch() here, or else Promise.reject() won't work.
```

The last point of discussions was whether a catch() block was even necessary for this particular lifecycle check, since we're unsure if the oAuth module passes such errors up/down the chain. 🤷‍♂️  If you _do_ want to catch oAuth errors however, you can make it _catch() first_ and then do a _then()_

Usable pattern (ver 2):
```
return oauth.checkBearerToken()
  .catch(error => // catch any errors thrown by the OAuth module)
  .then(token => // rejecting here will block the chain until the next catch block)
```

Reminder: if the `catch()` up there forgets to `return` anything, the `token` value in the `then()` will be `undefined`.

eatyourgreens:
I had a look through `oauth.js` and I honestly can't easily tell what happens if the network is down, or Panoptes is not responding. However, if your token can't refresh via the embedded iframe, it will resolve with the token set to `null`.

camallen:
You can simulate the network failure in chrome by setting "Offline" in network panel. 

You can simulate panoptes timing out by modifying the DNS resolution in `/etc/hosts` as outlined in https://docs.acquia.com/article/using-etchosts-file-custom-domains-during-development. That won't simulate gateway timeouts or other HTTP errors though.

eatyourgreens:
@camallen Cheers, but what does the `oauth.js` code actually do when the network is offline? That's confusing me.

shaunanoordin:
@wgranger This PR has been updated with:
- Improved API client calls
- `panoptes-client` 2.9.4
- Added Rollbar logging
- Emergency save now trumps manual save

simoneduca:
Thanks for the very useful summary, Shaun. I had tested SW/oauth flow with the Chrome `offline` and the `oauth.checkBearerToken()` just returns `null` still, like Jim said above 
>if your token can't refresh via the embedded iframe, it will resolve with the token set to null.

wgranger:
This works much better due to the client update. Should we also insert 

`dispatch(emergencySaveWorkInProgress());`
`dispatch(toggleDialog(<DialogOfFailure />, false, true));`

at the catch of `saveClassificationInProgress()`? I think that might provide the layer of protection there we have been needing? Ex. if the app is offline and you try to submit, you aren't currently notified your classification failed to save. Although, you might want to double check the classification submits successfully if we decide to do this.

shaunanoordin:
@wgranger hot dang, you're right. My earlier tests didn't catch that because I only tried the "what if suddenly the WiFi died?" scenario with _full Classification submission,_ which was fully caught by the emergency queue system.

Save Progress, meanwhile, is a blackhole. OK, putting up a fix now.

ALSO ALSO, Panoptes Client is getting ready for a 2.9.5 update, and I feel that we should wait for that latest version to be released. 2.9.5 fixes an issue with the auto-refreshing of tokens, which would be very minor for most other projects, but will affect ASM like crazy due to the amount of time our users stay on the page. (The bug is in 2.9.4, but not visible unless you were due for a token refresh, IIRC.)

eatyourgreens:
> (The bug is in 2.9.4, but not visible unless you were due for a token refresh, IIRC.)
That bug also only shows up if you have third-party cookies blocked, which is why I think we missed it before.

shaunanoordin:
## PR Update
- Now, when Save Progress fails, the emergency save safety net deploys.
- Also, I fixed an invisible bug - previously, when `emergencyLoadWorkInProgress()` triggered, the Subject ID was sometimes null (due to the way that fetchSavedSubject() and prepareForNewSubject() trigger; more details below).
  - Hence, a new function to manually specify Subject ID has been specified.

### So what was that invisible error?
Observe, in `emergencyLoadWorkInProgress()`:
```
dispatch(fetchSavedSubject(subjectId));
prepareForNewSubject(dispatch, null);
```

See, `fetchSavedSubject(subjectId)` ONLY records the value of `subjectid` when an asynchronous call to `apiClient.type('subjects')` succeeds. This means the value of Subject ID might get set AFTER prepareForNewSubject() runs, meaning prepareForNewSubject() might create a Classification with Subject ID = null.

Fortunately, inserting a manual update to the Subject ID in between those two calls fixes the issue. True, the manually set Subject ID will be overwritten by fetchSavedSubject() when the async call finishes, but it's basically the same value.

(Further dev rambling: Another approach is to set the Subject ID before any async calls begin, e.g. put it inside fetchSavedSubject(), before apiClient.type('subjects').get(id). This is fairly irrelevant for Save Progress continuation since, Save Progress retrieves a whole existing Classification (which already has its own Subject ID) instead of creating a new one (which requires that Subject ID be known))

### Status
Waiting for PJC 2.9.5

shaunanoordin:
## PR Update
PJC 2.9.5 is GO! Big thanks to Jim for sorting out the client code. 😄 

### Status
@wgranger ready for final review; once (if) this gets a 👍 and a merge, I'll deploy it to production.

src/ducks/subject.js
```diff
@@ -241,9 +261,12 @@ const fetchSubject = (initialFetch = false) => {
         });
     };
 
-    if (savedClassificationPrompt) {
+    if (checkEmergencySave()) {  //If not, check if there's an emergency save.
```
wgranger: How do we know the emergency save belongs to the user that is currently classifying? It is possible to create an emergency classification with a logged in user, sign out on the main page, then enter the classifier and continue that classification as a logged out user. I think the emergency save local storage item should be identified by the current user as the saved classification prompt currently is.

wgranger:
This is **very** close to being merged, but there is just one item that needs to be sorted out before a merge can happen.

eatyourgreens:
Not a blocking comment, but I get the impression the emergency save data doesn’t need to persist beyond the current tab, so could use `sessionStorage` instead of `localStorage`.

eatyourgreens:
There was a problem with `panoptes-client@2.9.5` but 2.9.6 should be good.

shaunanoordin:
## PR Update
This PR has been updated:
- We now use Panoptes-client 2.9.6
- The Emergency Save now cares whether a user is logged in when saving/retrieving emergency save data.
  - This means that after getting booted out by the Emergency Save prompt, a user must re-login with the _same Zooniverse account_ they used when they first encountered the transcription-crashing error.
  - See #254 for full details, especially for a consensus over the discussion of whether the safety net should check for user logins or not.

### Status
@wgranger Ready for re-review

src/components/DialogOfContinuation.jsx
```diff
@@ -0,0 +1,43 @@
...
+          onClick={(e) => {
+            dispatch(clearEmergencySave());
+            location.reload();
+          }}
+        >
+          New Manuscript
+        </button>
+        <button
+          className="button"
+          onClick={(e) => { onClose && onClose(e); }}
```
wgranger: I think after the annotation is loaded when clicking this button this should also `clearEmergencySave()`, otherwise the following might happen:

1) User makes some annotations and saves classification successfully
2) User makes some annotations, saves classification and it fails, emergency save kicks in
3) User reloads page and loads emergency classification
4) User makes some annotations and saves classification successfully
5) User reloads the page, but the emergency classification is again loaded from point number 3 above (losing any annotations at point 4)

shaunanoordin: This is a fair point.

Previously, I _purposely_ did not clear the emergency save _until_ the Classification was submitted, because there was a scenario where:
1. a user did work
2. user was logged out, then reloaded the page,
3. user returned to the page (but was logged out), quite possibly resuming their work,
4. then went "whoops! I need to login again", did so, and...
5. returned to the page with the emergency save data removed.

With the recent update for this branch (via PR #254), a user _has_ to be logged in to resume the work they did, so it's now fair to clear the emergency save when that happens. The manual Save Progress then takes the responsibility for maintaining user progress.

shaunanoordin:
Thanks for all the feedback; this PR has been updated further. 👍 

@wgranger I've added the clearEmergencySave() where you recommended in the DialogOfContinuation. Your suggestion makes sense - the recent inclusion of user-login checking means the code doesn't need to worry about losing the emergency save data when the user "steps out" to re-log in.

@eatyourgreens I've noted your suggestion on using sessionStorage, but I feel that the persistence of localStorage offers more safety to our users - I want the safety net to be as broad as possible, even if it means weirdness if a user, for whatever reason, opens the project on two windows. 🤔 

(Although a user _accidentally closing a tab_ should be a major _whoopsie_ that ASM needs to guard against - autosaves or 'you sure you want to close this window?' prompts might need to be the next set of safety features.)

wgranger:
I think this PR is in a good place. Note: we are also committing the package.lock with this branch.

PR #254, opened 2018-03-07
Optional for PR 251 - Emergency Save with login checks

shaunanoordin:
## PR Overview
This is an optional PR - it changes PR #251 so that the Emergency Save system cares whether a user is logged in or not.

### Status
Optional - do not merge into 251 until we can confirm this is the kind of app behaviour that we want; but if we do go for it, please remember to merge this before 251 is.

Tagging @wgranger

### DIFF
--- a/src/ducks/subject.js
+++ b/src/ducks/subject.js
@@ -261,9 +261,9 @@ const fetchSubject = (initialFetch = false) => {
         });
     };
 
-    if (checkEmergencySave()) {  //If not, check if there's an emergency save.
+    if (checkEmergencySave(getState().login.user)) {  //If not, check if there's an emergency save.
       dispatch(emergencyLoadWorkInProgress());
-      dispatch(toggleDialog(<DialogOfContinuation />, false, false));
+      dispatch(toggleDialog(<DialogOfContinuation dispatch={dispatch} />, false, false));
     } else if (savedClassificationPrompt) {  //Check if the user has manually saved progress. (Emergency save trumps manual save.)
       dispatch(toggleDialog(<ClassificationPrompt />, false, true));
     } else if (!getState().subject.queue.length) {  //If not, check if there are any subjects left in the queue.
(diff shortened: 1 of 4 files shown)


src/ducks/subject.js
```diff
@@ -261,9 +261,9 @@ const fetchSubject = (initialFetch = false) => {
         });
     };
 
-    if (checkEmergencySave()) {  //If not, check if there's an emergency save.
+    if (checkEmergencySave(getState().login.user)) {  //If not, check if there's an emergency save.
       dispatch(emergencyLoadWorkInProgress());
-      dispatch(toggleDialog(<DialogOfContinuation />, false, false));
+      dispatch(toggleDialog(<DialogOfContinuation dispatch={dispatch} />, false, false));
```
shaunanoordin: 😕

Not sure if adding the `dispatch` like this is "proper", but that's how it works.

shaunanoordin:
Based on our conversations on Slack, we're moving forward with this update to PR #251 

Arguments in favour of implementing login checks for the emergency save (i.e. this PR):
- Sam: I think it would be *really* confusing for someone to show up and see a half-finished letter.
- Will: Yeah, my worry was that someone would be working on a classification that wasn’t theirs. I think it should only work for the logged in user who initially began the classification.
- Becky: it would be a great incentive to make people log in
- Zach: tempted to agree, expectations of that should rely on being logged in, which solves the shared computer problem // yup, “log in to make sure your work isn’t lost!”

Arguments in favour of keeping the emergency save agnostic to the login state:
- Shaun: the emergency save is the 'final safety net' that should kick in when all else fails (e.g. the login system messing up) and shouldn't care who did the work, only that the work is saved.

There is a stronger support for login checks, so we're going with that. But please note that, moving forward with ASM development, we need to be aware of the weakness: if the fundamental login systems fail (e.g. any unexpected weirdness with the PJC), _nobody_ will be able to retrieve their lost work. Logically, this fail scenario _shouldn't_ happen, but I've seen enough glitches to worry about any possibility no matter how improbable.

### Status
Merging to #251

PR #257, opened 2018-03-14
getSessionID() should compare integer representations of Date objects

simensta:
While adding the get-session-id.js file to Ancient Lives, I noticed in `#getSessionID()` that `ttl < Date.now()` never returned true.
https://github.com/zooniverse/anti-slavery-manuscripts/blob/6b3aee3c51cb26273b20664bb218a946581000d4/src/lib/get-session-id.js#L25-L26
The method is comparing a string representation of a Date object with a number representation of a Date object. This PR makes a small adjustment so that integer value representations are compared instead.

### DIFF
--- a/src/lib/get-session-id.js
+++ b/src/lib/get-session-id.js
@@ -22,7 +22,7 @@ const getSessionID = () => {
     ({id, ttl} = JSON.parse(sessionStorage.getItem('session_id')));
   }
 
-  if (ttl < Date.now()) {
+  if (new Date(ttl).getTime() < Date.now()) {
     id = generateSessionID();
   } else {
     ttl = fiveMinutesFromNow();
--- a/src/lib/get-session-id.js
+++ b/src/lib/get-session-id.js
@@ -4,7 +4,7 @@ const generateSessionID = () => {
   const hash = require('hash.js');
   const sha2 = hash.sha256();
   const id = sha2.update(`${Math.random() * 10000 }${Date.now()}${Math.random() * 1000}`).digest('hex');
-  const ttl = fiveMinutesFromNow();
+  const ttl = twoHoursFromNow();
   const stored = {id, ttl};
   try {
     sessionStorage.setItem('session_id', JSON.stringify(stored));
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Whaaa! 😱 So we've been generating new sessions every time? This may also affect Scribes.

simensta:
I think the session id is suppose to indicate whether classifications have been submitted in the same 5 minute interval. However, when #getSessionID() is called in ducks/classification.js, the session id never changes/is not updated because `ttl < Date.now()` always returns false.
https://github.com/zooniverse/anti-slavery-manuscripts/blob/38b3186658b5156903886ff4ace37ef666d529f4/src/ducks/classifications.js#L311

eatyourgreens:
The session ID is supposed to change if there’s no activity for 5 minutes.

shaunanoordin:
Correction: so we _haven't_ been generating a new session ID all this time? 😱 

andrea++ and jim++ for the clarification.

OK, and now for the...

shaunanoordin:
Correction: so we _haven't_ been generating a new session ID all this time? 😱 

andrea++ and jim++ for the clarification.

OK, and now for the...

eatyourgreens:
The code does look like you’re expected to call `getSessionID()` regularly if you want to refresh the ttl. If you’re only calling it once for the classification, and classification takes longer than five minutes, then each classification would end up being a new session. 

I think this works in PFE because there’s an assumption that each classification takes less than five minutes.

wgranger:
Nice catch. Yes, I think it should be increased definitely. I think a new, solo classification could take an hour or two, but I don't think a collaborative classification would take quite as long. I'm inclined to say an hour or two?

eatyourgreens:
If someone transcribes a few lines, takes a break, then finishes the page, is that one session or two? If two, you should call `getSessionID()` after each line and expire after five minutes.

eatyourgreens:
Oh wait, that doesn’t matter because the session is only recorded on the classification. Ignore me.

wgranger:
Are we good to merge this? Not sure where we landed on the `fiveMinutesFromNow()` conversation.

shaunanoordin:
Thanks for the reminder, @wgranger 

I've updated the code to have a "two hours from now" function to replace "five minutes from now", though I feel like I should revisit this setup again later.

### Status
Merging. 👍

PR #262, opened 2018-07-10
Allow users to select Workflow 

shaunanoordin:
## PR Overview
This PR is a major change to the way the ASM project operates:
- users are now able to **choose** between the Solo and Collaborative workflows, when starting a project. 
- additionally, prep work is being done for a _future update_ where only one workflow is selected by default.

### Status
WIP. Tagging @snblickhan since she's directing this feature, and @wgranger for bonus dev eyes. (There's so much rewiring going on here.)

Done so far:
- when visiting the Classify page, the user will see a menu prompting them to choose either the Solo or Collaborative workflows.
  - This page can be expanded with more text, so we can explain to the user what their choices are.
  - Note: the user will **not** be able to select another workflow until they refresh the page. I could add this functionality in later, but right now it's a bit of a UI faff.
- basic prep work for "single workflow" future update.
- the Splits logic is being slowly stripped away.
- some refactoring of the Workflow duck to follow current duck standards.

Still needs work:
- 💀 Proper interaction with the Save/Resume Progress feature, which will be the biggest challenge, honestly.
- The UI design. AAAHHH @beckyrother AVERT YOUR EYES NOOO HISSSS
  screen shot 2018-07-10 at 17 15 10

PR #263, opened 2018-07-16
Remove Seven-Ten and Splits

shaunanoordin:
## PR Overview
This PR removes Seven-Ten and most of the splits logic (e.g. fetching and logging A-B split decisions), except for the variant Solo/Collab toggle (which is still used by the UI to determine which UI variant to show the user).

### Status
@wgranger this is ready for review.

NOTE: This is being merged into PR #262 , so should be reviewed before that one.


src/ducks/splits.js
```diff
@@ -1,18 +1,9 @@
-import { Split } from 'seven-ten';
 import { config } from '../config';
```
wgranger: `config` is no longer used in this component.

wgranger:
I think this PR is good to go. I just have one request with moving the `config` import from a component.

shaunanoordin:
## PR Update
Thanks for spotting that, Will! The unnecessary `config` element has now been removed from the Splits duck. I'll now merge this into #262

PR #264, opened 2018-07-27
Update markdownz to 7.6.3

eatyourgreens:
Fixes a github security warning for the string dependency in older versions of markdownz.
Updates the webpack config because two of the markdown plugins are now written in ES6, so should be passed through babel.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
       "dev": true,
       "requires": {
-        "mime-types": "2.1.18",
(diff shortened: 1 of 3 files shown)


wgranger:
Good heads up. Thanks for the update.

eatyourgreens:
I've just noticed that this PR changed a load of unrelated packages in the lock file. I'm not sure what happened there.

wgranger:
Yes. I assumed you did a fresh npm install when updating the package and it updated the items in the lock files.

eatyourgreens:
Looks like it mostly downgraded modules to older versions. Weird.

PR #265, opened 2018-08-10
Updated packages after running npm audit fix

srallen:
Ran `npm audit fix` after getting a security warning about this package still using an old version of superagent. We updated superagent in panoptes-javascript-client, so this project was a few minor versions behind with PJC. I didn't test this app at all, but the update went fine with PFE.


wgranger:
Looks good. Thanks for the keen eye.

PR #267, opened 2018-09-04
Update Button Text and Add Announcement Banner

wgranger:
This PR comes on the heels of #262 and makes two changes.

- Adjusts the button text from "solo" to ~~"collaborative"~~ "independent" for the workflow selection button to keep naming consistent. 

- Reintroduces the announcement banner to the top of ASM notifying users of the feedback form.

### DIFF
--- a/src/components/Banner.jsx
+++ b/src/components/Banner.jsx
@@ -2,20 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const Banner = ({ hideBanner }) => {
-  //If no messages, just return null.
-  return null;
-  
-  /*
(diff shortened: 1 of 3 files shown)


shaunanoordin:
- Wait, you mean you renamed "Solo" to "Independent", correct?
- You also need to change the text in the Classifier toolbar that says "You
are currently working on {currentMode}
- - Cmd+F and find `const currentMode`
- - I actually renamed the display name of the "Individual" variant to
"Solo" in that component, ahhah.

wgranger:
Ah yes, good eye. I can make that change now.

shaunanoordin:
👍

PR #275, opened 2018-09-25
Fix missing "Resume Progress" popup - don't render app until login is initialised

### LINKED ISSUE #274: "Resume Progress" popup sometimes goes missing
## Bug
When a user has Saved Work In Progress, and attempts to resume it, they sometimes don't see the "We found your saved work" popup that prompts them to resume/cancel their work in progress.

screen shot 2018-09-25 at 16 34 07

Based on this thread: https://www.zooniverse.org/projects/bostonpubliclibrary/anti-slavery-manuscripts/talk/1183/738154?page=3

### Analysis

The situation:
- You open https://www.antislaverymanuscripts.org/classify
- You login, or are already logged in, into User Account A
- You make transcriptions to Subject X, either in Solo or Collab mode
- You choose Save Progress
- You close the browser and/or computer window
- You come back the next day to https://www.antislaverymanuscripts.org/classify
- You login or are already logged in as User A
- **You expect to see a popup that says "We Found Your Saved Work" (with prompts to resume/cancel your work in progress for Subject X) but instead do NOT see that popup**

The absence of the popup indicates an anomaly stemming from the Saved Progress Check block in ClassifierContainer.jsx.

However, we can already confirm (see the Talk thread) that the bug affects users who are logged in and using the same computer, so for now let's rule out the "cookies have mysteriously vanished" theory. Let's assume that 

Reviewing the Resume Saved Progress check, the code looks for...

```
if (this.props.user && localStorage.getItem(`${this.props.user.id}.manual_save_classificationID`))
``` 

Hypothesis: `this.props.user` isn't registering properly - i.e. **the Classifier page doesn't know that the user is supposed to be already logged in when it's checking for saved work** 

Testing method:
- Log in, go to the Classify page, do a transcription, Save Progress.
- Reload the page.
  - Expected (if hypothesis is true): the Resume Progress should appear _most of the time._ (Not necessarily all)
  - Perform several times to confirm.
- **Add a small 2s network lag ** (using dev tools, longer delays acceptable)
  - Expected (if hypothesis is true): the Resume Progress does NOT appear _most of the time._

Testing this on antislaverymanuscripts.org has given me strong indications that the hypothesis is correct, meaning that the we're encountering **a race condition.**

**The Classifier page is likely checking to see if a user has any Saved Work before even checking to see if there's a user logged in**

The apparent random rate that the users have reported encountering this problem seems consistent with how we know random network lag may affect when the app receives confirmation on the user logged in state.

### Suggested Solution

Theoretically this should be simple to fix - we prevent the app from even rendering unless we know for sure that the user's logged in state can be determined (i.e. in App.jsx, if `!props.initialised`, show Loading circle.)

### Status
Testing a solution

shaunanoordin:
## PR Overview
This PR fixes an issue where the Resume Progress popup doesn't show, even if the user has Saved Work In Progress (as shown in the browser cookies). Turns out the problem is that sometimes the Classifier page attempts to find a user's work in progress before realising if there's a user to begin with.

See #274 for more details.

Fixes #274 

In any case, this initialisation check is good practise for _all our projects_ as the whole _point_ of the `.initialised` parameter is to tell the app that it's ready to render (at least in terms of user login checks).

Note to self: check out projects to ensure that this is also in place.

### Status
@wgranger this is ready for review. If everything's OK, can you also please help deploy the fix? I need to head off shortly-ish.

Tagging @snblickhan so we can update our volunteers later.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -97,7 +97,7 @@ class App extends React.Component {
   }
 
   render() {
-    if (this.props.projectStatus !== PROJECT_STATUS.READY) {
+    if (this.props.projectStatus !== PROJECT_STATUS.READY || !this.props.initialised) {
       return <LoadingSpinner />
     }  //TODO: Consider what to do for STATUS: ERROR
 


wgranger:
Hard to believe that this huge issue was a one-line fix. I've tested this locally and the saved work popup is appearing as expected. Merging now. After the merge, I'll deploy master.

PR #277, opened 2018-10-16
Make easier to retire

### LINKED ISSUE #276: Items not greying out when needed
Looks like a check was never added for greyed lines when the number of views is greater than five. Logic should be added (where grey line logic exists) to check for:

`(annotation.consensus_score >= 3) || (annotation.number_views >= 5)`

shaunanoordin:
## PR Overview
This PR makes it easier for lines of text to be considered "retired", thus helping Subjects get retired.

As @CKrawczyk pointed out, the front end was checking that the consensus score exceeded a certain amount, but did not actually check if a line was already viewed a certain number of lines (to indicate that this is probably a line that's too difficult to get a consensus on, so should be retired anyway).

Closes #276 

### Status
Ready for review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
       "dev": true,
       "requires": {
-        "mime-types": "2.1.18",
+        "mime-types": "~2.1.16",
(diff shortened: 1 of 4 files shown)


wgranger:
LGTM

Thanks for the quick fix.

PR #278, opened 2018-10-25
Disable Collaborative Option

wgranger:
This PR disables the collaborative option from the workflow selection window. This will be put in place temporarily until new subjects are uploaded.

screen shot 2018-10-25 at 10 47 39 am


wgranger:
I'm merging this and deploying now as this is an important and timely fix.

PR #281, opened 2018-11-12
Remove Workflow Selection

wgranger:
Do not merge until we get the green light from the BPL team that the field guide has been updated and new subjects added.

This PR removes workflow selection from the classifier. I _think_ all our bases are covered here; however, with all the code to check for variants, it would be good to have a sharp eye on this. In a nutshell, changes here:

- Set default variant to `collaborative`
- Remove workflow selection section

Most of the code referencing variants and independent mode has been left in should we choose to revert back to that variant at some point.

### DIFF
--- a/src/components/SelectedAnnotation.jsx
+++ b/src/components/SelectedAnnotation.jsx
@@ -75,8 +75,8 @@ class SelectedAnnotation extends React.Component {
     let textAfter;
     let textInBetween;
 
-    const startTag = '[' + textTag + ']';
-    const endTag = '[/' + textTag + ']';
+    const startTag = `[${textTag}]`;
+    const endTag = `[/${textTag}]`;
     const text = this.inputText;
     const textAreaValue = text.value;
     const selectionStart = text.selectionStart;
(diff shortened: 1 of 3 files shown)


shaunanoordin:
There are some merge conflicts here, so let me resolve them.

shaunanoordin:
## PR Review
This PR switches the logic of ASM to _always_ use the **Collaborative Workflow.** The ability to choose between two workflows (as in the previous phase) or the ability to be auto-sorted to one workflow (as in the original phase) has been removed altogether.

@wgranger please be aware that I rebased this branch and made one small update, so you'll want to refresh your local branch.

### Status
LGTM, but holding off on the merge until green light is given.

wgranger:
This should be good to merge on Monday. Note: I added a table text modifier, but it operates the same as other modifiers. Would be good to double check everything looks alright in production before a merge and deployment.

shaunanoordin:
## PR Update

The latest changes LGTM, though I can't really "see" the collaborations of Collaborative mode in Production because all the subjects - like the Prince of Bel Air - are just way too fresh.

Aside: I don't see any notes requesting that the _"This project is undergoing some light maintenance and will resume shortly, with new letters to transcribe!"_ should be removed, so I won't touch it for now. I can update this later.

### Status

Right, this is getting merged and deployed

PR #285, opened 2019-02-13
Improve Caesar Error Handling

### LINKED ISSUE #284: Handle Caesar Errors
Currently, too much of the site relies on successful responses of Caesar. Just checkout Rollbar. I believe many of the errors there are from the lack of existential checks to handle the absence of Caesar data. This would probably alleviate some of the issues volunteers are experiencing.

https://www.zooniverse.org/projects/bostonpubliclibrary/anti-slavery-manuscripts/talk/1183/879124?comment=1451830&page=1

wgranger:
Closes #284 

Users are reporting errors of seeing annotations from different subjects. This PR attempts to solve those issues by making several changes:

**Main changes:**
- Fixes a bug where users can submit a blank classification using keyboard shortcuts (ctrl + enter). Try selecting a previous annotation, keep the box blank, and hit the shortcut.
- Ensure subject id fetched from Caesar matches the displayed subject id.
- Fix error seen in Rollbar where user tries to update a non-existent annotation (probably from incorrectly shown lines of a previous annotation).
- Adds catches to all promises.

**Other changes:**
- Fixes various linter issues, mainly line indents

The main error received from Rollbar recently is an `unhandled rejection was null or undefined!` error stemming from [this call](https://github.com/zooniverse/anti-slavery-manuscripts/blob/master/src/components/App.jsx#L52) failing, which happens whenever a user's token expires. Take a look at the screenshot below, made after the addition of the catch in this PR. Both the `then` and `catch` are reached due to the `Promise.reject`. 

screen shot 2019-02-13 at 8 41 15 am

I could've added an exception to the Rollbar config to ignore unhandled rejections, but it seemed better practice to add catches.

**Note:** This will not necessarily _solve_ all errors received from Caesar, but it should help users experiencing annotations appearing from different subjects.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -51,21 +51,18 @@ class App extends React.Component {
 
     return oauth.checkBearerToken()
       .then((token) => {
-        //If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
+        // If the App thinks you're logged in, but the token says otherwise, deploy emergency measures.
         if (props.initialised && props.user && !token) {
           this.props.dispatch(emergencySaveWorkInProgress());
           this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
           return Promise.reject(new Error('User is supposed to be logged in, but token has expired.'));
-          //The intent is that if the user is supposed to be logged in but
-          //isn't, the whole request (that comes after .beforeEveryRequest)
-          //should not continue.
+          // The intent is that if the user is supposed to be logged in but
+          // isn't, the whole request (that comes after .beforeEveryRequest)
+          // should not continue.
         }
         return;
-      })
-      .catch((err) => {
-        console.error('Bearer Token Fetch Failed: ', err);
       });
-      //Note: do NOT add a catch() here, or else Promise.reject() above won't stop the API request.
+    // Note: do NOT add a catch() here, or else Promise.reject() above won't stop the API request.
   }
 
   getChildContext() {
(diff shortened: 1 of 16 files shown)


src/components/App.jsx
```diff
@@ -45,22 +45,25 @@ class App extends React.Component {
...
           this.props.dispatch(toggleDialog(<DialogOfFailure />, false, true));
           return Promise.reject(new Error('User is supposed to be logged in, but token has expired.'));
           //The intent is that if the user is supposed to be logged in but
           //isn't, the whole request (that comes after .beforeEveryRequest)
           //should not continue.
-        }      
+        }
         return;
+      })
+      .catch((err) => {
```
shaunanoordin: Wait, the line below literally says `//Note: do NOT add a catch() here, or else Promise.reject() above won't stop the API request.` - are we sure we want to do this?

shaunanoordin: I need to double check this, but if we add `.catch()` here, then if any other API call (e.g. fetch workflow, submit Classification) fails due to a bad bearer token, then those API calls' respective `.catch()`es won't trigger.

eatyourgreens: Yep. The error here means that your login session has expired and couldn't be renewed (probably because the browser or a privacy extension blocks zooniverse.org while you're on antislaverymanuscripts.org.)

I'd guess that catching that error means you'll lose your classification work?

wgranger: I failed to see that comment below, so I'm glad you're pointing it out to me now. That error has ceased being reported from Rollbar, so I'm ok removing the `catch` here if we're worried about adverse effects.

shaunanoordin:
## PR Update
Awesome, the latest changes fixes the concerns I was having about this PR. These changes now LGTM, and will be merged and deployed.

PR #287, opened 2019-04-05
Update Webpack

wgranger:
This PR updates to webpack 4 and babel 7. This should take care of any issues with IE experiencing syntax errors.

Note, this PR also contains the commit from #286, so it can be merged in lieu of that or rebased once that PR is merged.


shaunanoordin:
## PR Review

This PR 1. updates our Webpack code bundler _and_ 2. fixes an issue where our old Polyfill code wasn't actually filling any polys (the previous URL to the polyfill has 404ed) - both with the goal of fixing ASM for our _IE11_ users.

### Status

Testing in progress.

1. I have no issues with the Webpack updates as I'm able to run and build (`npm start` and `npm run build` with no issues)

2. the Polyfill service needs a more thorough test on my end (on Virtual box) - although I can confirm that the new URL (`https://polyfill.io/v3/polyfill.min.js?features=default,es6,Promise,Array.prototype.includes,Array.prototype.keys`) is functioning properly, there were still reports that some IE11 users continued to have issues.

shaunanoordin:
Update: Nope, the Polyfill still isn't covering everything that IE11 requires. Currently detected shortcoming: `=>` fat arrows for declaring functions.

I'm going to see if `https://polyfill.io/v3/polyfill.min.js` works better or if we need to spell out every feature required.

EDIT: wait, hang on, weren't JS function arrows introduced in ES6?

package.json
```diff
@@ -66,23 +76,26 @@
...
+    "html-webpack-plugin": "^3.2.0",
+    "image-webpack-loader": "~4.6.0",
     "json-loader": "^0.5.4",
-    "mocha": "^3.2.0",
+    "mini-css-extract-plugin": "^0.5.0",
+    "mocha": "~4.0.0",
     "nib": "^1.1.2",
     "publisssh": "^1.1.0",
     "react-addons-test-utils": "^15.4.2",
+    "react-hot-loader": "~3.0.0-beta.7",
```
srallen: I believe hot module functionality is built into webpack-dev-server, so I don't think you need this anymore: https://webpack.js.org/configuration/dev-server/#devserverhot

wgranger:
Yes, there was an issue with the markdown modules not compiling arrow functions correctly, which is why `exclude: /node_modules\/(?!(markdown-it-anchor|markdown-it-table-of-contents|striptags)\/).*/,` was added, but perhaps there is still setup required to account for that.

shaunanoordin:
Wait. Oh no. Can the `=>` arrow even be polyfilled? To an older browser, that's basically a new/illegal _syntax_ isn't it?

If so then the question I need to ask is, "Why do we have => in our code? Shouldn't I bundlers/transpilers be doing something about it?" Checking now.

shaunanoordin:
> Yes, there was an issue with the markdown modules not compiling arrow functions correctly, which is why exclude: /node_modules\/(?!(markdown-it-anchor|markdown-it-table-of-contents|striptags)\/).*/, was added, but perhaps there is still setup required to account for that.

Ah, so this has been noted then! But I'm still seeing `const slugify = (s) => encodeURIComponent(String(s).trim().toLowerCase().replace(/\s+/g, '-'))` coming from the `markdown-it-anchor` . Let me check if that `exclude` line has been applied to both production and development, and if it's triggering properly.

shaunanoordin:
## PR Review (Update)

OK, cool! I managed to get ASM running on IE11 - specifically, on IE11+OS Virtualbox.

The combination of `webpack: exclude: /node_modules\/(?!(markdown-it-anchor|markdown-it-table-of-contents|striptags)\/).*/,` and the new `Polyfill.io` path solves the issue where the ASM website wouldn't even render in IE11.

### (Minor) Change Requested

- [ ] `exclude: /node_modules\/(?!(markdown-it-anchor|markdown-it-table-of-contents|striptags)\/).*/,` should be added to `webpack.config`

While the `webpack-exclude` fix is technically only needed in `webpack.production.config` (since IE11 users would only be accessing the production website, not dev/staging), it might still be a good idea to keep our webpack configs consistent.

### Status

This PR LGTM, though I imagine we can only _really_ confirm the fix is working once we've deployed to production. Goddarnit, IE11. 😢

wgranger:
Although approved, let me make a few changes to this based on recent comments before merge.

wgranger:
Changes made. Feel free to take another look and merge when ready.

shaunanoordin:
These updates LGTM 👍 I've reinstalled all dependencies and I can build fine.

I'll proceed to merge this PR and deploy shortly.

PR #291, opened 2019-06-12
[DO NOT MERGE] Update ZRC and run `npm audit fix` - Mk2

shaunanoordin:
## PR Review

The intent of this PR is to re-do #289, except this time make it work when deployed to `https://www.antislaverymanuscripts.org/`. ⚠️ So far, I have not been successful. 😢 

Changes in this PR:
- Updated Zooniverse dependencies, as per #289 
- Ran `npm run audit`, as per #289 
- Removed the pointless `npm run test` and `npm run test-travis` scripts, which weren't really doing anything.
- Project now requires the use of Node v10

Unfortunately, I'm hitting the same snags that forced me to revert with #290 
- the ASM website looks fine on localhost, with `npm start` on OSX+Chrome
- However, when I run `npm run deploy-production`, the live website simply stops working. Blank page, no error messages in the console.
- ❓🤷‍♂❓ 

Observations:
- When I run `npm run deploy` I see only two files being updated: `index.html` and `vendor-chunkhash.js`. Hmm...
- To test that the compiled files look fine, I ran `npm run build`, then ran a simple `express` server in the created `/dist` directory. The website still looks fine.

### Status

DO NOT MERGE. This is currently an investigation PR.

@wgranger I'd like to request a second set of eyes on this, if you have the time. CCing @srallen for updates

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -939,6 +939,21 @@
         }
       }
     },
+    "@babel/runtime": {
+      "version": "7.4.5",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.4.5.tgz",
+      "integrity": "sha512-TuI4qpWZP6lGOGIuGWtp9sPluqYICmbk8T/1vpSysqJxRPkudh/ofFWyqdcMsDf2s7KvDL4/YHgKyvcS3g9CJQ==",
+      "requires": {
+        "regenerator-runtime": "^0.13.2"
+      },
+      "dependencies": {
+        "regenerator-runtime": {
+          "version": "0.13.2",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.2.tgz",
+          "integrity": "sha512-S/TQAZJO+D3m9xeN1WTI8dLKBBiRgXBlTJvbWjCThHWZj9EvHK70Ff50/tYj2J/fvBY6JtFVwRuazHN2E7M9BA=="
+        }
+      }
+    },
     "@babel/template": {
       "version": "7.4.0",
       "resolved": "https://registry.npmjs.org/@babel/template/-/template-7.4.0.tgz",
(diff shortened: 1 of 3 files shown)


wgranger:
I think looking at the webpack production config is a good first thing to look at, but it wouldn't necessarily explain why you were able to build the current master back in April (since there are no changes here to the config). 

However, I did notice there is a difference with the output `filename` and `chunkFilename` on this app compared to what [we have in PFE](https://github.com/zooniverse/Panoptes-Front-End/blob/master/webpack.build.js#L34).

shaunanoordin:
Thanks to Will's testing, we can confirm that the breaking change is ZRC moving from `0.7.0` to `0.7.3`. We'll look into this more later.

shaunanoordin:
Next steps: I need to talk to @adammcmaster about creating a Staging setup for `anti-slavery-manuscripts`, because the error we're seeing only happens on the live production servers instead of on local. (Hi Adam! I'll tell you more on Slack tomorrow)

srallen:
@shaunanoordin `publisssh` will create a bucket folder for you if the specific location in the command does not already exist, so I believe you can just add an npm script that does something like:

`"deploy-staging": "NODE_ENV=production npm run build && publisssh dist zooniverse-static/preview.zooniverse.org/asm-staging"`

Then navigating to `https://preview.zooniverse.org/asm-staging` should work.

I would just confirm with @adammcmaster that `preview.zooniverse.org` is still being used for this. 

If you add `-d` flag to the publisssh command, it'll do a dry run too.

shaunanoordin:
## PR Update

Derp! I deployed this branch safely on Monday 17 June after the problem was identified and resolved, but forgot to merge this branch. Anyway, recap:

Problem: when publishing updated version of ASM to `antislaverymanuscripts.org`, the website stops working.

Diagonsis: this was due to a combination of two problems.
- the `index.html` was cached on the server side
- the `vendor` JS bundle was named `vendor-chunkhash.js` (instead of `vendor-1234abcd-random-hash.js`) due to a typo in webpack.config

This resulted in the `vendor` JS bundle being updated but the `main` (app) JS bundle - both being referred to by the cached `index.html` being out of sync with one another. i.e. one bundle would be a few code revisions ahead/behind the other.

Solution: clearing the server-side cache, easily done via `lita` commands on Slack. (See `#dev` conversation on Fri 14  June)

### Status
Merging

PR #303, opened 2020-02-28
Create Staging Script

wgranger:
This PR creates a staging script so we can get a staging version deployed. This will make it easier for other users to log in and submit staging classifications through the CFE.

**Use case:**
We need to test out the ALICE pipeline, but we'd like to try it with a fresh subject set. I've created a subject set on ASM staging for this purpose, but I'd like multiple people to submit classifications. Rather than get multiple non-devs set up with the local app, etc. etc. I think it'd be easier to get a staging site deployed, finally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test-travis": "true",
     "eslint": "eslint .",
     "build": "BABEL_ENV=production check-engines && check-dependencies && webpack --config webpack.production.config.js -p",
-    "_stage": "BABEL_ENV=staging check-engines && check-dependencies && webpack --config webpack.config.js -p",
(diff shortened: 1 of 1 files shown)


wgranger:
Just ran this script, which appears to be working, but I think I need to adjust the `oauth` redirect uri as I'm seeing this when I try logging on
Screen Shot 2020-03-02 at 3 12 34 PM

wgranger:
This should be good after adding 'antislaverymanuscripts.preview.zooniverse.org' to Doorkeeper

wgranger:
Confirmed: I can login via the deployed staging site.

PR #306, opened 2020-06-10
Remove Rollbar error reporting system

shaunanoordin:
## PR Overview

This PR removes the [Rollbar](https://rollbar.com/) error reporting system from ASM.

Some context:

- Rollbar was first introduced when we were trying to figure out what would be the "user session timeout" error. Now that we've identified it, we haven't been using the error logs too often, but it's still handy to have.
- We eventually want to replace Rollbar with [Sentry](https://sentry.io/)
- First things first though, we'll remove Rollbar since ASM is eating into the error reporting quota like mad.

@wgranger notes:

- When we eventually want to re-introduce error tracking, the insertion points are easy to find: we just Cmd+F to find `console.error` and `console.info`, plus whatever initialisation setup is necessary. 👍 

### Status

Ready for review.


shaunanoordin:
👍 Cool, merging today - but deploying on Monday, because let us not risk a Friday deploy.

